PR #3, opened 2017-05-21
Initial framework

vibhavp:
The `wasm` package should be able to read a WebAssembly binary (the optional name section isn't parsed, however).


sbinet:
@vibhavp it's a lot of code to review in one go...
Would you mind to chop it off in more piece wise PRs?

Also, as a more general comment: it would be great to start with a documented API :)

sbinet:
(and apologies for the belated answer...)

vibhavp:
good idea, I'll see if I can break it into multiple PRs.

PR #4, opened 2017-05-26
wagon: add initial doc + gofmt test

sbinet:
This CL adds an initial documentation placeholder for the wagon package.
It also adds a simple test to make sure files are being gofmt'ed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,8 @@
+language: go
+go:
+  - 1.6.x
+  - 1.7.x
+  - 1.8.x
+  - tip
+
+sudo: false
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -0,0 +1,8 @@
+language: go
+go:
+  - 1.6.x
+  - 1.7.x
+  - 1.8.x
+  - tip
```
shurcooL: FWIW, Travis refers to `tip` as `master` now. See https://docs.travis-ci.com/user/languages/go/#Specifying-a-Go-version-to-use. `tip` still works, but it's unclear if it'll stop working in the future.

sbinet: thanks. fixed.

PR #6, opened 2017-06-06
Initial code.

vibhavp:


dennwc:
Added few comments to current version.

What do you think about adding some sample wasm modules to serve as test cases?

---

Reviewed 29 of 29 files at r1, 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 31 unresolved discussions.

---

*[cmd/readwasm/readwasm.go, line 9 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-Klz26Eahk3Rx1IKEW-l:-Klz26Eahk3Rx1IKEW-m:b-uytvgx) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/cmd/readwasm/readwasm.go#L9)):*
> ```Go
> 	"github.com/go-interpreter/wagon/validate"
> 	"github.com/go-interpreter/wagon/wasm"
> 	flag "github.com/ogier/pflag"
> ```

`github.com/spf13/pflag` should be more up-to-date

---

*[validate/error.go, line 20 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-Klz2Y2RVYvdGZW08od9:-Klz2Y2RVYvdGZW08odA:bz89b19) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/validate/error.go#L20)):*
> ```Go
> 
> type ValidateError struct {
> 	Offset int
> ```

Is this a PC or a byte offset? From `mockVM.NewValidateError` I would have said it's a PC, but from the struct definition it looks like a bytes offset.

---

*[validate/vm.go, line 91 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-Klz7WUzhI_0eb8FjksK:-Klz7WUzhI_0eb8FjksL:be8e7h2) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/validate/vm.go#L91)):*
> ```Go
> 
> func (vm *mockVM) topOperand() (o operand, under bool) {
> 	stackTop := vm.stackTop - 1
> ```

Maybe use `len(vm.stack)` instead of separate variable?

---

*[wasm/imports.go, line 19 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-Klz9QR6Z_S4_tDLx6Hu:-Klz9QR6Z_S4_tDLx6Hv:bn2s4gm) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/imports.go#L19)):*
> ```Go
> 	// If Kind is Memory, Type is a Memory containing the type of the imported memory
> 	// If the Kind is Global, Type is a GlobalVar
> 	Type interface{}
> ```

It can be replaced with type enum to force compiler to check type on this field. In this case Kind may also be omitted in favor of Go type system:

```golang
// Import is a type enumeration implemented by:
// - FuncImport
// - TableImport
// - MemoryImport
// - GlobalImport
type Import interface{
  isImport() // ensures that only this package can implement it
}

type FuncImport struct{
  Type uint32
}
func (FuncImport) isImport() {}

type TableImport struct{
  Type Table
}
func (TableImport) isImport() {}

type MemoryImport struct{
  Type Memory
}
func (MemoryImport) isImport() {}

type GlobalImport struct{
  Type GlobalVar
}
func (GlobalImport) isImport() {}
```

This way it will preserve type safety and prevent some potential bugs.

---

*[wasm/imports.go, line 40 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzBe0t95o1KsMqa1WZ:-KlzBe0uhxlCX1UO5Jao:b-isolyq) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/imports.go#L40)):*
> ```Go
> 		importedModule, ok := modules[importEntry.ModuleStr]
> 		if !ok {
> 			rdr, err := resolve(importEntry.ModuleStr)
> ```

Make ResolveFunc return a decoded module instead? If module is imported by multiple other modules, it will reuse existing decoded version.

---

*[wasm/imports.go, line 86 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzBs5lcH49M_akwuM9:-KlzBs5lcH49M_akwuMA:bidfqag) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/imports.go#L86)):*
> ```Go
> 			}
> 		case ExternalTable:
> ```

TODO/panic/error; also add `default` case

---

*[wasm/index.go, line 61 at r2](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzOqPTku1xP4Enm-Yl:-KlzOqPTku1xP4Enm-Ym:bbde2wp) ([raw file](https://github.com/go-interpreter/wagon/blob/ba6ea3dfab82d044a3a78f6214448733d3b8d5c4/wasm/index.go#L61)):*
> ```Go
> func (m *Module) populateTables() error {
> 	if m.Table == nil || len(m.Table.Entries) == 0 || m.Elements == nil || len(m.Elements.Entries) == 0 {
> 		return nil
> ```

Something in this condition might require throwing an error. Or not?

---

*[wasm/init_expr.go, line 9 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzDAKtYwWXrnfrjQkA:-KlzDAKtYwWXrnfrjQkB:b-3rqco4) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/init_expr.go#L9)):*
> ```Go
> const end byte = 0x0b
> 
> func readInitExpr(r io.Reader) ([]byte, error) {
> ```

Change type to `io.ByteReader`?

---

*[wasm/init_expr.go, line 31 at r2](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzP2YdZ4luD3iyjBm9:-KlzP2YdZ4luD3iyjBmA:bbnhfhk) ([raw file](https://github.com/go-interpreter/wagon/blob/ba6ea3dfab82d044a3a78f6214448733d3b8d5c4/wasm/init_expr.go#L31)):*
> ```Go
> // either be int32, int64, float32 or float64
> // It returns an error if the initializer expression is inavlid.
> func (m *Module) ExecInitExpr(expr []byte) (interface{}, error) {
> ```

If `type Expr []byte` is added, this can be a method on it.

---

*[wasm/module.go, line 18 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzDnA8UIq4oFXz-fq-:-KlzDnA8UIq4oFXz-fq0:b-974lad) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/module.go#L18)):*
> ```Go
> 
> type Module struct {
> 	Magic   uint32
> ```

Any reason to save magic here?

---

*[wasm/module.go, line 69 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzEz6MEyt-V0_NX7UP:-KlzEz6MEyt-V0_NX7UQ:b94is53) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/module.go#L69)):*
> ```Go
> 		done, err := m.readSection(reader)
> 		if err != nil {
> 			log.Fatal(err)
> ```

`return` instead

---

*[wasm/module.go, line 81 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzF86XwIdExbU1kGln:-KlzF86XwIdExbU1kGlo:bbs37tx) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/module.go#L81)):*
> ```Go
> 		return nil, err
> 	}
> ```

TODO to run other populate functions?

---

*[wasm/section.go, line 37 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzF_RULyuRqF45MhJw:-KlzF_RULyuRqF45MhJx:bwzu15) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/section.go#L37)):*
> ```Go
> 	ID SectionID
> 	// Size of this section in bytes
> 	PayloadLen uint32
> ```

Rename to `Size`? Or even make a function to calculate this from `End-Start`.

---

*[wasm/section.go, line 87 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzHwPjkY3lx4g-Rxo6:-KlzHwPjkY3lx4g-Rxo7:blj4twh) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/section.go#L87)):*
> ```Go
> 	s.Start = r.CurPos
> 
> 	s.Bytes = make([]byte, payloadDataLen)
> ```

This copy can be omitted (section can be read directly from stream), but this can wait until VM is finished, I guess.

---

*[wasm/section.go, line 88 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzHKVxTOjCTyV9vM9U:-KlzHKVxTOjCTyV9vM9V:bxvl14t) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/section.go#L88)):*
> ```Go
> 
> 	s.Bytes = make([]byte, payloadDataLen)
> 	_, err = r.Read(s.Bytes)
> ```

Payload might be >32K, and Read provides no guarantee that it will read full buffer of this size in one call. Use `io.ReadFull` instead. 

---

*[wasm/section.go, line 188 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzIYDF8Qb7TboU5JtV:-KlzIYDF8Qb7TboU5JtW:btn8m4t) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/section.go#L188)):*
> ```Go
> 	}
> 
> 	m.Type = s
> ```

Please name this field `Types`. It's confusing to see that `Module.Type` is set to some section.

---

*[wasm/section.go, line 403 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzJLZVnu7e3GwPa9w_:-KlzJLZWMbFRcm8bDV-D:b-qz7xed) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/section.go#L403)):*
> ```Go
> 
> 	for i := uint32(0); i < count; i++ {
> 		var err error
> ```

No need to declare it, it will be initialized below

---

*[wasm/section.go, line 488 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzJcgbmjhiPKcYKPks:-KlzJcgbmjhiPKcYKPkt:bj6obwl) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/section.go#L488)):*
> ```Go
> 	Index uint32
> 	// should return i32, computes the offset at which to place the elements
> 	Offset []byte
> ```

Maybe add `type Code []byte` and/or `type ExprCode []byte` definition and use it here and in other places? It will be useful to attach methods to it.

---

*[wasm/section.go, line 582 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzLDKh3v6mkZveBaCr:-KlzLDKiNRL7FINGqaxI:b-n1ailm) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/section.go#L582)):*
> ```Go
> 	}
> 
> 	r = io.LimitReader(r, int64(bodySize))
> ```

It might be a good idea to drain this reader in two failure cases below. This way decoder can continue it's work, marking one function body as invalid.

---

*[wasm/section.go, line 600 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzKWqDiQlKw21GwIFn:-KlzKWqDiQlKw21GwIFo:b32n2gk) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/section.go#L600)):*
> ```Go
> 	code, err := ioutil.ReadAll(r)
> 
> 	if code[len(code)-1] != 0x0b {
> ```

There might be a case when this "end" exists in the middle of code section, and then more data follows (malformed code). Will validation catch this case?

---

*[wasm/types.go, line 77 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzM3B8053OdVYm3BYd:-KlzM3B9O1w0NXyi3vEv:bj9p081) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/types.go#L77)):*
> ```Go
> 	ParamTypes []ValueType
> 	// Number of results from the function
> 	ReturnCount uint32
> ```

Any reason to save it here? In the future it will be `len(ReturnTypes)`, and for now it's okay to check `ReturnType != nil`, I think.

---

*[wasm/types.go, line 191 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzMUAHwqdZuUENEkgw:-KlzMUAI2rmwIfa9mPQu:b-lhbeup) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/types.go#L191)):*
> ```Go
> 
> type Memory struct {
> 	Limits *ResizableLimits
> ```

This field should not be optional, correct?

---

*[wasm/types.go, line 231 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzMreafqYRT2qCX_HY:-KlzMreafqYRT2qCX_HZ:b-pxkc98) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/types.go#L231)):*
> ```Go
> type ResizableLimits struct {
> 	// true if the maximum field is present
> 	Flags bool
> ```

Maybe drop it, since there is a pointer on `Maximum`? Or make an `int`. It looks redundant currently.

---

*[wasm/types.go, line 235 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzMhQENKkmTVQQD3b9:-KlzMhQENKkmTVQQD3bA:boeiu64) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/types.go#L235)):*
> ```Go
> 	Initial uint32
> 	// non-nil if specified by flags
> 	Maximum *uint32
> ```

It's okay to assume that `0 == not set` and omit pointer.

---

*[wasm/types.go, line 247 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzN6_rGs0qOeN6pUkI:-KlzN6_rGs0qOeN6pUkJ:b68feg9) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/types.go#L247)):*
> ```Go
> 	}
> 
> 	lim.Flags = f == 1
> ```

There might be more flags later. Use`f & 0x1 != 0` instead.

---

*[wasm/internal/readpos/readpos.go, line 7 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzH56AK_RNCL-kVM3c:-KlzH56AK_RNCL-kVM3d:b803oey) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/internal/readpos/readpos.go#L7)):*
> ```Go
> )
> 
> type ReadPos struct {
> ```

Move to `read.go`?

---

*[wasm/internal/readpos/readpos.go, line 14 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzGXIptOCZPKuwFtOz:-KlzGXIptOCZPKuwFtP-:bpkx437) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/internal/readpos/readpos.go#L14)):*
> ```Go
> func (r *ReadPos) Read(p []byte) (int, error) {
> 	n, err := r.R.Read(p)
> 	if err == nil {
> ```

It's better to omit this check. Some readers may return `n > 0` and `err = io.EOF`. And there is no harm in making this counter to increment on error.

---

*[wasm/leb128/read_test.go, line 17 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzNYig5Er2VwPNqWyu:-KlzNYig5Er2VwPNqWyv:b-vxsaaj) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/leb128/read_test.go#L17)):*
> ```Go
> 	sn, err := ReadVarint(bytes.NewReader([]byte{0xFF, 0x7e}))
> 	assert.NoError(t, err)
> 	assert.Equal(t, sn, int32(-129))
> ```

Add tests with padding?

---

*[wasm/operators/const.go, line 8 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzO07lB4ziykfIEFjt:-KlzO07lB4ziykfIEFju:b-yoi4fb) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/operators/const.go#L8)):*
> ```Go
> 
> var (
> 	I32Const = newOp(0x41, "i32.const", []wasm.ValueType{}, wasm.ValueTypeI32)
> ```

Just `nil` instead of `[]wasm.ValueType{}`?

---

*[wasm/operators/op.go, line 11 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzOIa8fl63IsC-WpBp:-KlzOIa8fl63IsC-WpBq:b-mb453t) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/operators/op.go#L11)):*
> ```Go
> var (
> 	ops      = make([]*Op, 256)
> 	noReturn = wasm.ValueType(wasm.BlockTypeEmpty)
> ```

`const`?

---

*[wasm/operators/op.go, line 51 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzOTmb55KGP3vEFltQ:-KlzOTmb55KGP3vEFltR:b-65onie) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/operators/op.go#L51)):*
> ```Go
> }
> 
> func DecodeOp(code byte) (*Op, error) {
> ```

Maybe `GetOp` is better? It doesn't decode anything.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-:-KlzPKBNgkub7ue2LWSv:b-j3nl6g)*

sbinet:
ok, here is what I propose.

as I said earlier, we need to break this huge "code drop" into manageable pieces :).

one possible way is to do:

```
$> git remote -v
origin	https://github.com/go-interpreter/wagon (fetch)
origin	https://github.com/go-interpreter/wagon (push)
vibhavp	git://github.com/vibhavp/wagon.git (fetch)
vibhavp	git://github.com/vibhavp/wagon.git (push)

$> git fetch --all -p
$> git checkout -b gsoc ecf1e6f
$> git log
* 1142f29 (vibhavp/master) Rename ExecInitExpr -> execInitExpr
* b5fcfbe Rewrite readInitExpr, complete execInitExpr.
* 47f40c5 populateGlobals: Log the number of entries in the global index space.
* 5edf64e Handle offsets for element segments correctly, add populateLinearMemory.
* a6c2fb9 SectionExports: Remove unused field `Count'
* bf3a2e4 ReadModule: Populate global and table index spaces.
* a32d64f Rename field (wasm.Module).Type to Types.
* cd0932d ReadModule: Return error from readSection
* ba6ea3d Complete populateTables.
* b4ca6ab Add ExecInitExpr (not implemented yet).
* f73d0c9 Add basic logger that can be disabled.
* 0da5eac Add staging code for population index spaces of a module.
* 141e7e4 Add staging code for resolving imports
* 3896303 Add package validate.
* 6078cfc Add StackTypeError
* c64c2f5 Add 64 bit variants.
* bdcf6d2 ReadModule: Add resolver function argument.
* 072479b Add Start and End fields to sections, add logging.
* f88d7d8 Implement Stringer for FunctionSig and BlockType values.
* 64d8106 readInitExpr: Dont use bufio.
* 8134590 Make readpos internal
* a6ee795 Validate modules
* f8d2e06 Print start and end addresses for sections
* e088e27 Add stack params to operator values.
* 36781e2 Remove unneeded error
* 142b271 Add binary 'readwasm'
* 129447d Add package wasm
| * ecf1e6f (HEAD -> gsoc, origin/master, origin/HEAD, master) wagon: travis-ci renamed tip into master
| | * e746d49 (origin/contribute-doc) wagon: add contribute guide
| |/  
| * c465c30 wagon: add godoc+travis badges
| * 950a016 wagon: add initial travis-ci config
| * de0a5a5 wagon: add initial doc + gofmt test
|/  
* 21c4a5b doc: typo
* d0bb38c wagon: first import
```

at this point, you can "easily" copy some files from your `master` branch into this new `gsoc` branch, where we'll land the alluded to piecewise PRs.

e.g.: copy the `wasm/leb128` package into the current `gsoc` branch:

```sh
$> ls -l
total 16K
-rw-r--r-- 1 binet binet  254 Jun  7 09:38 doc.go
-rw-r--r-- 1 binet binet 1.6K May 29 11:54 LICENSE
-rw-r--r-- 1 binet binet  391 Jun  7 09:38 README.md
-rw-r--r-- 1 binet binet  799 Jun  7 09:38 wagon_test.go

$> mkdir -p wasm/leb128
$> git show vibhavp/master:wasm/leb128/read.go > wasm/leb128/read.go 
$> git status
On branch gsoc
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	wasm/

nothing added to commit but untracked files present (use "git add" to track)

$> git add -p wasm/leb128/read.go
On branch gsoc
Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

	new file:   wasm/leb128/read.go

$> git checkout vibhavp/master wasm/leb128/read_test.go 
$> git status
On branch gsoc
Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

	new file:   wasm/leb128/read.go
	new file:   wasm/leb128/read_test.go

$> git commit -m "wasm/leb128: initial import"
$> git push vibhavp gsoc
```

I'd propose to start with `wasm/leb128` (because it's the package with the least dependencies) and work our way up (`wasm/internal/readpos`, `wasm`, `wasm/operators`)

here is the package dependency list:

```sh
$> for i in `find ./wasm -type d`; do echo "
> === $i ===";
> deplist $i; done

=== ./wasm ===
github.com/go-interpreter/wagon/wasm
github.com/go-interpreter/wagon/wasm/internal/readpos
github.com/go-interpreter/wagon/wasm/leb128

=== ./wasm/leb128 ===
github.com/go-interpreter/wagon/wasm/leb128

=== ./wasm/operators ===
github.com/go-interpreter/wagon/wasm
github.com/go-interpreter/wagon/wasm/internal/readpos
github.com/go-interpreter/wagon/wasm/leb128
github.com/go-interpreter/wagon/wasm/operators

=== ./wasm/internal ===
github.com/go-interpreter/wagon/wasm/internal

=== ./wasm/internal/readpos ===
github.com/go-interpreter/wagon/wasm/internal/readpos

=== ./validate ===
github.com/go-interpreter/wagon/validate
github.com/go-interpreter/wagon/wasm
github.com/go-interpreter/wagon/wasm/internal/readpos
github.com/go-interpreter/wagon/wasm/leb128
github.com/go-interpreter/wagon/wasm/operators
```

please make sure to follow the contribution guidelines (not yet official/enforced, but as nobody yelled yet...) :
 https://github.com/go-interpreter/license/pull/1/commits/0161e16829eb3c129a16b1be55bcc54384373a40

notably: please add a license header to the new files, and document a bit more (at least) the exported functions.

vibhavp:
Thanks. I'll do this if I once I address @dennwc's review comments.

sbinet:
SGTM

sbinet:
Review status: 23 of 29 files reviewed at latest revision, 29 unresolved discussions.

---

*[wasm/module.go, line 12 at r4](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KmAmIgDeWX6sFQE-VWL:-KmAmIgDeWX6sFQE-VWM:bszyvrs) ([raw file](https://github.com/go-interpreter/wagon/blob/1142f2987cfb023fecd3199cfaaca7a7b789b9b7/wasm/module.go#L12)):*
> ```Go
> 
> const (
> 	WasmMagic   uint32 = 0x6d736100
> ```

as it's in the `wasm` package, I would just call it `Magic`.
same for `WasmVersion` below.

---

*[wasm/section.go, line 599 at r4](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KmAnHyDfOFituDqyL8v:-KmAnHyDfOFituDqyL8w:b-o7hd0v) ([raw file](https://github.com/go-interpreter/wagon/blob/1142f2987cfb023fecd3199cfaaca7a7b789b9b7/wasm/section.go#L599)):*
> ```Go
> 	code, err := ioutil.ReadAll(r)
> 
> 	if code[len(code)-1] != 0x0b {
> ```

also, perhaps use the `end` const defined somewhere instead of a `0x0b` literal.
not that the value would change at this point in time, but it's marginally clearer :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/go-interpreter/wagon/6)*

sbinet:
Review status: 23 of 29 files reviewed at latest revision, 32 unresolved discussions.

---

*[cmd/readwasm/readwasm.go, line 9 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-Klz26Eahk3Rx1IKEW-l:-KmAssdGtwq0nYN3XP3G:bzczmh4) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/cmd/readwasm/readwasm.go#L9)):*

`github.com/spf13/pflag` should be more up-to-date
</blockquote></details>

is this external dependency really warranted?
OK, `"flag"` doesn't have all the bells and whistles of GNU's `getopt` but it's not like `readwasm` has gazillions of command line options.

I would start with just `"flag"` and if people complain and insist on having all the features of `"pflag"`, then migrate to that.

---

*[wasm/leb128/read_test.go, line 7 at r4](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KmAtVZtloJjRgGaeXkF:-KmAtVZtloJjRgGaeXkG:bdpuhc7) ([raw file](https://github.com/go-interpreter/wagon/blob/1142f2987cfb023fecd3199cfaaca7a7b789b9b7/wasm/leb128/read_test.go#L7)):*
> ```Go
> 	"testing"
> 
> 	"github.com/stretchr/testify/assert"
> ```

I think we should stick with using the standard lib's `"testing"`.
ok, `testify/assert` is marginally terser on the page, but using `"testing"` (and plain `if/else/return`) looks more like regular Go code.

---

*[wasm/operators/op.go, line 1 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KmArH1YzXqqyAG33ofP:-KmArH1YzXqqyAG33ofQ:bh3dj6y) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/operators/op.go#L1)):*
> ```Go
> package operators
> ```

maybe just name it `ops` (or `wasmops`? `wops`?<)  instead of `operators` ?
you do import it as `ops` somewhere else.

actually, wouldn't this make more sense to merge it inside `wasm` instead?

---

*[wasm/operators/op.go, line 45 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KmAsQ8W9ZdWoVmDfflD:-KmAsQ8W9ZdWoVmDfflE:b-twb6ba) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/operators/op.go#L45)):*
> ```Go
> }
> 
> type InvalidOpcodeError byte
> ```

idiomatic Go is to name errors with `ErrSomething`.
_e.g._ 
```
$> go doc io ErrNoProgress
$> go doc io ErrUnexpectedEOF
```

thus, I would rename it `ErrInvalidOpcode`

---

*[wasm/operators/op.go, line 51 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzOTmb55KGP3vEFltQ:-KmAsCvfDdbRIANZvs9Z:b-d58mj5) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/operators/op.go#L51)):*

Maybe `GetOp` is better? It doesn't decode anything.
</blockquote></details>

or `NewOpFrom` or `OpFrom`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/go-interpreter/wagon/6)*

dennwc:
Reviewed 5 of 5 files at r3, 2 of 2 files at r4.
Review status: all files reviewed at latest revision, 33 unresolved discussions.

---

*[wasm/init_expr.go, line 38 at r4](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KmBD_DDT9tQ-FovSvbI:-KmBD_DDT9tQ-FovSvbJ:b-mlvr0u) ([raw file](https://github.com/go-interpreter/wagon/blob/1142f2987cfb023fecd3199cfaaca7a7b789b9b7/wasm/init_expr.go#L38)):*
> ```Go
> 				return []byte{}, err
> 			}
> 			binary.Write(buf, binary.LittleEndian, i)
> ```

I was wondering if it's necessary to re-encode the expression this way. Maybe this should just decode and validate the expression without changing format, and then `execInitExpr` will read the same format? It might be useful, because we can call `exec` to run an expression directly and verify it in-flight.

---

*[wasm/leb128/read_test.go, line 7 at r4](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KmAtVZtloJjRgGaeXkF:-KmBCJXO7TpgEW2PnSHy:b-a8n9g0) ([raw file](https://github.com/go-interpreter/wagon/blob/1142f2987cfb023fecd3199cfaaca7a7b789b9b7/wasm/leb128/read_test.go#L7)):*

I think we should stick with using the standard lib's `"testing"`.
ok, `testify/assert` is marginally terser on the page, but using `"testing"` (and plain `if/else/return`) looks more like regular Go code.
</blockquote></details>

I would say that code with `testify` is more readable when `testing`. The code has no branching, thus it become easier to read the sequence of operations in the test.

Also, it's not a dependency of the library itself. User will only need to install it if he want's to run tests. And even in this case there is a high chance that it will be installed already - `testify` is used by a significant number of projects.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/go-interpreter/wagon/6)*

sbinet:
Review status: all files reviewed at latest revision, 32 unresolved discussions.

---

*[wasm/leb128/read_test.go, line 7 at r4](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KmAtVZtloJjRgGaeXkF:-KmBHYl33CIEOGhd9XdU:bwlgmx2) ([raw file](https://github.com/go-interpreter/wagon/blob/1142f2987cfb023fecd3199cfaaca7a7b789b9b7/wasm/leb128/read_test.go#L7)):*

I would say that code with `testify` is more readable when `testing`. The code has no branching, thus it become easier to read the sequence of operations in the test.

Also, it's not a dependency of the library itself. User will only need to install it if he want's to run tests. And even in this case there is a high chance that it will be installed already - `testify` is used by a significant number of projects.
</blockquote></details>

it's annoying to remember that one has to `go get -t github.com/go-interpreter/wagon/...` to also get the testing dependencies.

we used to use `testify/assert` in Gonum and gonum/plot but reverted to just `testing`.

I'd propose to first start with just `testing` and revisit later on when we have more tests.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/go-interpreter/wagon/6)*

vibhavp:
*[validate/vm.go, line 91 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-Klz7WUzhI_0eb8FjksK:-KmGjrhHM1anvWhfU_KN:bop18qn) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/validate/vm.go#L91)):*

Maybe use `len(vm.stack)` instead of separate variable?
</blockquote></details>

Because of the structure of control blocks in wasm, each block has it's own stack. Instead of using a new stack  every time we encountered a new block, I thought it would be better idea to store the current stack top value at the beginning of a new stack and restore it whenever we get out of it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/go-interpreter/wagon/6)*

vibhavp:
*[wasm/imports.go, line 40 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzBe0t95o1KsMqa1WZ:-KmFT-mX7Re7Zorh6AQQ:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/imports.go#L40)):*

Make ResolveFunc return a decoded module instead? If module is imported by multiple other modules, it will reuse existing decoded version.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/go-interpreter/wagon/6)*

vibhavp:
Review status: 16 of 30 files reviewed at latest revision, 34 unresolved discussions.

---

*[wasm/imports.go, line 86 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzBs5lcH49M_akwuM9:-KmFT2LFQ6OrUWoK7s8z:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/imports.go#L86)):*

TODO/panic/error; also add `default` case
</blockquote></details>

Done.

---

*[wasm/module.go, line 81 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzF86XwIdExbU1kGln:-Km6UaumHTz04Z9MTeMO:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/module.go#L81)):*

TODO to run other populate functions?
</blockquote></details>

Done.

---

*[wasm/module.go, line 12 at r4](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KmAmIgDeWX6sFQE-VWL:-KmFT62mkAzDk8l-B7Pe:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/1142f2987cfb023fecd3199cfaaca7a7b789b9b7/wasm/module.go#L12)):*

as it's in the `wasm` package, I would just call it `Magic`.
same for `WasmVersion` below.
</blockquote></details>

Done.

---

*[wasm/section.go, line 188 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzIYDF8Qb7TboU5JtV:-Km6Ujthb7AENRa8Uq3b:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/section.go#L188)):*

Please name this field `Types`. It's confusing to see that `Module.Type` is set to some section.
</blockquote></details>

Done.

---

*[wasm/section.go, line 403 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzJLZVnu7e3GwPa9w_:-Km6UlpDfWc_Z1J9VnNv:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/section.go#L403)):*

No need to declare it, it will be initialized below
</blockquote></details>

Done.

---

*[wasm/types.go, line 77 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzM3B8053OdVYm3BYd:-KmFTAiT175b8sR7cVOx:besfhvh) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/types.go#L77)):*

Any reason to save it here? In the future it will be `len(ReturnTypes)`, and for now it's okay to check `ReturnType != nil`, I think.
</blockquote></details>

I've changed all of this to a single `ReturnTypes` slice field

---

*[wasm/internal/readpos/readpos.go, line 14 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KlzGXIptOCZPKuwFtOz:-KmFTYna4AuwO_m8rQn4:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/f73d0c9e92aa130152e26cc3a46d696fe03c8487/wasm/internal/readpos/readpos.go#L14)):*

It's better to omit this check. Some readers may return `n > 0` and `err = io.EOF`. And there is no harm in making this counter to increment on error.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/go-interpreter/wagon/6)*

dennwc:
Reviewed 14 of 14 files at r5.
Review status: all files reviewed at latest revision, 27 unresolved discussions.

---

*[validate/error.go, line 74 at r5](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KmHNXCX9JTt7b6x6OYS:-KmHNXCX9JTt7b6x6OYT:bkvh4ob) ([raw file](https://github.com/go-interpreter/wagon/blob/61c083606e1e03240d9adbc7891669ea43b7985e/validate/error.go#L74)):*
> ```Go
> }
> 
> type InvalidGlobalIndexError struct {
> ```

I should be `ErrInvalidGlobalIndex`, according to Go convention. Please change other errors accordingly.

---

*[wasm/index.go, line 111 at r5](https://reviewable.io:443/reviews/go-interpreter/wagon/6#-KmHJ-pgKjBY5kif55mR:-KmHJ-pgKjBY5kif55mS:b-og618e) ([raw file](https://github.com/go-interpreter/wagon/blob/61c083606e1e03240d9adbc7891669ea43b7985e/wasm/index.go#L111)):*
> ```Go
> // by the integer index. It returns an error if index is invalid.
> func (m *Module) GetTableElement(index int) (uint32, error) {
> 	if index >= len(m.TableIndexSpace[0]) {
> ```

check if `index < 0` as well; same for other similar functions

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/go-interpreter/wagon/6)*

techtonik:
What happened with this PR?

sbinet:
It was just too big.
It has been split into more manageable pieces.

PR #8, opened 2017-06-16
wasm: New package

vibhavp:
* wasm is the central package for reading and parsing WASM modules


wasm/types.go
```diff
@@ -0,0 +1,273 @@
...
+	return External(bytes[0]), err
+}
+
+type ResizableLimits struct {
+	// true if the maximum field is present
+	Flags bool
+	// initial length (in units of table elements or wasm pages)
+	Initial uint32
+	// non-nil if specified by flags
+	Maximum *uint32
```
sbinet: having a non-nil pointer to some value *and* a boolean flag that encodes whether that pointer is non-nil feels a bit redundant.

either loose the `Flags` and keep the `Maximum` field as a `*uint32`, or,
keep `Flags` but have the `Maximum` field be a `uint32` (which has the nice side-effect to reduce allocation).

what do you think?

wasm/types.go
```diff
@@ -0,0 +1,273 @@
...
+		return "<unknown external_kind>"
+	}
+}
+func readExternal(r io.Reader) (External, error) {
+	bytes, err := readBytes(r, 1)
+
+	return External(bytes[0]), err
+}
+
+type ResizableLimits struct {
```
sbinet: perhaps document a bit more what `ResizableLimits` is?

wasm/section.go
```diff
@@ -0,0 +1,696 @@
...
+
+type ErrInvalidSectionID SectionID
+
+func (e ErrInvalidSectionID) Error() string {
+	return fmt.Sprintf("wasm: invalid section ID %d", e)
+}
+
+var ErrUnsupportedSection = errors.New("wasm: unsupported section")
+
+func (m *Module) readSection(r *readpos.ReadPos) (bool, error) {
```
sbinet: please document what the boolean means.

sbinet:
@dennwc do you want to give it a try as well ?

wasm/imports.go
```diff
@@ -0,0 +1,104 @@
+package wasm
+
+import (
+	"errors"
+	"fmt"
+)
+
+// ImportEntry describes an import statement in a Wasm module.
+type ImportEntry struct {
+	// module name string
```
vibhavp: Yeah, I was just following the names given by the spec. `Name` would be a better prefix.

wasm/types.go
vibhavp: What would be the standard approach to this in idiomatic Go code?

wasm/init_expr.go
```diff
@@ -0,0 +1,171 @@
...
+	// integers into their normal little endian representation
+	// One reason why we do not execute it on the fly is that
+	// get_global uses indices to the global index space, which
+	// might have not been populated when a function reading a module
+	// section is calling this.
+outer:
+	for {
+		_, err := io.ReadFull(reader, b)
+		if err != nil {
+			return []byte{}, err
```
vibhavp: Oops, didn't know `nil` slices were a thing.

wasm/section.go
```diff
@@ -0,0 +1,696 @@
...
+	case ExternalTable:
+		logger.Println("importing table")
+		i.Type, err = readTable(r)
+	case ExternalMemory:
+		logger.Println("importing memory")
+		i.Type, err = readMemory(r)
+	case ExternalGlobal:
+		logger.Println("importing global var")
+		ptr, err = readGlobalVar(r)
+		if ptr != nil {
```
vibhavp: Thanks, forgot that case. That would end up panicking due to nil dereference.

vibhavp: From `readGlobalVar`'s code, that shouldn't happen. Anyway, I've added a better check for `err != nil` instead.

wasm/section.go
```diff
@@ -0,0 +1,696 @@
...
+		}
+	}
+
+	m.Data = s
+	return err
+}
+
+type DataSegment struct {
+	Index uint32
+	// should return i32, computes the offset at which to place the elements
```
vibhavp: Yes, that's due to the encoding for [`elem_segment`](https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#element-section) and [`data_segment`]([url](https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#data-section)) being the same.

wasm/types.go
sbinet: I *think* that:

```go
type ResizableLimits struct {
    Flags bool
    Initial uint32
    Maximum uint32
}
```
would be the most idiomatic Go code.

but perhaps we could just loose `Flags` altogether?
what is a `ResizableLimits` value with `Maximum == 0`  ?

vibhavp: > but perhaps Flags will also encode some additional fields of ResizableLimits

Yes, you're right. @dennwc had pointed that out in another review, so this will have to stay.

wasm/init_expr.go
```diff
@@ -0,0 +1,175 @@
...
+		case getGlobal:
+			index, err := leb128.ReadVarUint32(r)
+			if err != nil {
+				return nil, err
+			}
+			binary.Write(buf, binary.LittleEndian, index)
+		case end:
+			break outer
+		default:
+			return nil, errors.New("Invalid opecode in initializer expression")
```
sbinet: `s/opecode/opcode/`

wasm/index.go
```diff
@@ -0,0 +1,175 @@
...
+			fallthrough
+		case int(offset) == len(table):
+			table = append(table, elem.Elems...)
+		case int(offset) < len(table):
+			lastIndex := int(offset) + len(elem.Elems) - 1
+			if lastIndex >= len(table) {
+				table = append(table, make([]uint32, lastIndex+1-len(table))...)
+			}
+
+			tableIndex := offset
```
sbinet: wouldn't this be simpler rewritten as:

```go
for i, fct := range elem.Elems {
    table[i+offset] = fct
}
```
?

vibhavp: Yep, done.

dennwc:
Reviewed 1 of 9 files at r2, 1 of 2 files at r5, 7 of 8 files at r6, 1 of 1 files at r7, 1 of 1 files at r8.
Review status: all files reviewed at latest revision, 72 unresolved discussions.

---

*[wasm/imports.go, line 22 at r8](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn5AiRDag9PQw1rSRSc:-Kn5AiRDag9PQw1rSRSd:bn2s4gm) ([raw file](https://github.com/go-interpreter/wagon/blob/d51097470837f3534c28ed859303afcaee27a59f/wasm/imports.go#L22)):*
> ```Go
> 	// If Kind is Memory, Type is a Memory containing the type of the imported memory
> 	// If the Kind is Global, Type is a GlobalVar
> 	Type interface{}
> ```

It can be replaced with type enum to force compiler to check type on this field. In this case Kind may also be omitted in favor of Go type system:

```golang
// Import is a type enumeration implemented by:
// - FuncImport
// - TableImport
// - MemoryImport
// - GlobalImport
type Import interface{
  isImport() // ensures that only this package can implement it
}

type FuncImport struct{
  Type uint32
}
func (FuncImport) isImport() {}

type TableImport struct{
  Type Table
}
func (TableImport) isImport() {}

type MemoryImport struct{
  Type Memory
}
func (MemoryImport) isImport() {}

type GlobalImport struct{
  Type GlobalVar
}
func (GlobalImport) isImport() {}
```

This way it will preserve type safety and prevent some potential bugs.

---

*[wasm/imports.go, line 58 at r8](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn5B8pmRr07oOu5iGZE:-Kn5B8pmRr07oOu5iGZF:b-2ufgtk) ([raw file](https://github.com/go-interpreter/wagon/blob/d51097470837f3534c28ed859303afcaee27a59f/wasm/imports.go#L58)):*
> ```Go
> 		exportEntry, ok := importedModule.Export.Entries[importEntry.FieldName]
> 		if !ok {
> 			return errors.New("Couldn't import field.")
> ```

Specify field name in error

---

*[wasm/index.go, line 19 at r8](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn5BbncdS0MpOXcoRTP:-Kn5BbncdS0MpOXcoRTQ:b-f26a2d) ([raw file](https://github.com/go-interpreter/wagon/blob/d51097470837f3534c28ed859303afcaee27a59f/wasm/index.go#L19)):*
> ```Go
> 	}
> 
> 	if m.Function != nil {
> ```

Invert condition and place it as "or" condition to an if above

---

*[wasm/index.go, line 95 at r8](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn5CElBsqm7GMtVa4EP:-Kn5CElBsqm7GMtVa4EQ:b-gjf76o) ([raw file](https://github.com/go-interpreter/wagon/blob/d51097470837f3534c28ed859303afcaee27a59f/wasm/index.go#L95)):*
> ```Go
> 			lastIndex := int(offset) + len(elem.Elems) - 1
> 			if lastIndex >= len(table) {
> 				table = append(table, make([]uint32, lastIndex+1-len(table))...)
> ```

This looks unclear at a first glance. How about reallocating slice with a new size and `copy` an old one to it?

---

*[wasm/index.go, line 150 at r8](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn5D4gidYkESinFlByY:-Kn5D4gidYkESinFlByZ:b-eun5t4) ([raw file](https://github.com/go-interpreter/wagon/blob/d51097470837f3534c28ed859303afcaee27a59f/wasm/index.go#L150)):*
> ```Go
> 			lastIndex := int(offset) + len(entry.Data) - 1
> 			if lastIndex >= len(memory) {
> 				memory = append(memory, make([]byte, lastIndex+1-len(memory))...)
> ```

Same as above - allocate with a new size instead of doing index calculations to grow the slice. Later it will make sense to scan all linear memory indexes to get a maximal offset+size, allocate once an copy everything there.

---

*[wasm/index.go, line 153 at r8](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn5CvR8BJZsv4q1Wd1D:-Kn5CvR8BJZsv4q1Wd1E:byehn0g) ([raw file](https://github.com/go-interpreter/wagon/blob/d51097470837f3534c28ed859303afcaee27a59f/wasm/index.go#L153)):*
> ```Go
> 			}
> 
> 			for i, byte := range entry.Data {
> ```

Don't shadow builtins and use `copy` instead of a loop

---

*[wasm/init_expr.go, line 42 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVr-r1-42:-Kn5E5A5PuOYCO9mbELf:bdp8ra1) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/init_expr.go#L42)):*

is that really necessary?
couldn't you just write that one-element byte slice directly to `buf` ?
</blockquote></details>

Still unresolved. Write `b` directly to buf.

---

*[wasm/init_expr.go, line 49 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVr-r1-49:-Kn5EMAq9_Dhl_PdI09w:bhaul8g) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/init_expr.go#L49)):*

I don't have a handle on how performance-critical `readInitExpr` is, but `binary.Write` involves a bit of `reflect` code.
either use directly the `binary.LittleEndian.PutUint32` method or add a possible `FIXME` there.
</blockquote></details>

I agree here, it's better to use `binary.LittleEndian.PutUint32` directly, it will be much faster.

Also, I still don't understand why you re-encode the expression to a different format to execute it later.  If you want to have it in Buffer, you can make an `io.TeeReader`, and it will automatically populate the buffer while you reading this expression.

---

*[wasm/init_expr.go, line 107 at r8](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn5FH2No8zdvvrpHoE6:-Kn5FH2No8zdvvrpHoE7:b-wteipw) ([raw file](https://github.com/go-interpreter/wagon/blob/d51097470837f3534c28ed859303afcaee27a59f/wasm/init_expr.go#L107)):*
> ```Go
> 		b, err := r.ReadByte()
> 		if err != nil {
> 			if err == io.EOF {
> ```

```golang
if err == io.EOF {
  break
} else if err != nil {
  return nil, err
}
```

---

*[wasm/init_expr.go, line 115 at r8](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn5FUjmeLhDCDpBPmlk:-Kn5FUjmeLhDCDpBPmll:b5b6oew) ([raw file](https://github.com/go-interpreter/wagon/blob/d51097470837f3534c28ed859303afcaee27a59f/wasm/init_expr.go#L115)):*
> ```Go
> 		case i32Const:
> 			var i int32
> 			err := binary.Read(r, binary.LittleEndian, &i)
> ```

Same here - avoid `binary.Read`, it uses reflection which is slow.

---

*[wasm/module.go, line 22 at r8](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn5Ft07SJR1-Qo9iCR6:-Kn5Ft07SJR1-Qo9iCR7:b1legph) ([raw file](https://github.com/go-interpreter/wagon/blob/d51097470837f3534c28ed859303afcaee27a59f/wasm/module.go#L22)):*
> ```Go
> // Module represents a parsed WebAssembly module: http://webassembly.org/docs/modules/
> type Module struct {
> 	Magic   uint32
> ```

Any reason to save magic here? It will no be different from the constant because the code will throw an error in this case anyway.

---

*[wasm/section.go, line 97 at r8](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn5QAJoHhfuNZdJ7tuR:-Kn5QAJoHhfuNZdJ7tuS:b-7kvo11) ([raw file](https://github.com/go-interpreter/wagon/blob/d51097470837f3534c28ed859303afcaee27a59f/wasm/section.go#L97)):*
> ```Go
> 
> 	s.Bytes = make([]byte, payloadDataLen)
> 	_, err = io.ReadFull(r, s.Bytes)
> ```

Place a TODO to read it from stream directly without allocation the whole buffer.

---

*[wasm/section.go, line 603 at r8](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn5SYnfkGIEiJEfbdzJ:-Kn5SYnfkGIEiJEfbdzK:b5qo8si) ([raw file](https://github.com/go-interpreter/wagon/blob/d51097470837f3534c28ed859303afcaee27a59f/wasm/section.go#L603)):*
> ```Go
> 	}
> 
> 	r = io.LimitReader(r, int64(bodySize))
> ```

It might be a good idea to `defer io.Copy(ioutil.Discard, r)` to drain remaining body if it had some errors in it. In this case the next section might be read without errors and this function can be marked as damaged, for example.

---

*[wasm/section.go, line 621 at r8](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn5TBDnJ5Q9P8THNEp0:-Kn5TBDnJ5Q9P8THNEp1:b-bb7ykd) ([raw file](https://github.com/go-interpreter/wagon/blob/d51097470837f3534c28ed859303afcaee27a59f/wasm/section.go#L621)):*
> ```Go
> 	code, err := ioutil.ReadAll(r)
> 	logger.Printf("Read %d bytes for function body", len(code))
> 	if err != nil && err != io.EOF {
> ```

`ReadAll` never returns `EOF`

---

*[wasm/types.go, line 171 at r8](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn5Wx2oAXsG5eL1q9Gf:-Kn5Wx2oAXsG5eL1q9Gg:bwzjhit) ([raw file](https://github.com/go-interpreter/wagon/blob/d51097470837f3534c28ed859303afcaee27a59f/wasm/types.go#L171)):*
> ```Go
> 	}
> 
> 	g.Mutable = m == 1
> ```

This is a flag and it may be extended later. So use `m & 0x1 != 0` to select a specific bit.

---

*[wasm/types.go, line 266 at r8](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn5XDwAcTYgIQaa50Vb:-Kn5XDwAcTYgIQaa50Vc:b-gpjcyd) ([raw file](https://github.com/go-interpreter/wagon/blob/d51097470837f3534c28ed859303afcaee27a59f/wasm/types.go#L266)):*
> ```Go
> 	}
> 
> 	if lim.Flags&0x1 == 1 {
> ```

Later it will look like this: `flags & 0x32 == 0x32`. It's better to not duplicate constants in the expression: `flags & 0x32 != 0`. Same in case of `0x1`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/go-interpreter/wagon/8)*

wasm/imports.go
```diff
@@ -0,0 +1,104 @@
...
+	// If Kind is Function, Type is a uint32 containing the type index of the function signature
+	// If Kind is Table, Type is a Table containing the type of the imported table
+	// If Kind is Memory, Type is a Memory containing the type of the imported memory
+	// If the Kind is Global, Type is a GlobalVar
+	Type interface{}
+}
+
+var ImportMutGlobalError = errors.New("wasm: cannot import global mutable variable")
+
+type ErrInvalidExternal uint8
```
shurcooL: Reference: https://dmitri.shuralyov.com/idiomatic-go#error-variable-naming.

dennwc:
Reviewed 1 of 5 files at r9, 5 of 5 files at r10.
Review status: all files reviewed at latest revision, 71 unresolved discussions.

---

*[wasm/index.go, line 15 at r10](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn6Axru_I9f_pXYPJAy:-Kn6Axru_I9f_pXYPJAz:b-ruocp4) ([raw file](https://github.com/go-interpreter/wagon/blob/9b6b3bef1d94ca2b5051b2e768676ec854ca1bc9/wasm/index.go#L15)):*
> ```Go
> 
> func (e InvalidTableIndexError) Error() string {
> 	return fmt.Sprintf("wasm: Invalid table to table index space: %d", e)
> ```

It's better to include a cast here: `int(e)`. It's dangerously close to infinite recursion (`print` will try to call `e.Error` if format placeholder is different).

---

*[wasm/section.go, line 37 at r10](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn6BvrzoGs74qdmI4Ai:-Kn6BvrzoGs74qdmI4Aj:b-9dbm17) ([raw file](https://github.com/go-interpreter/wagon/blob/9b6b3bef1d94ca2b5051b2e768676ec854ca1bc9/wasm/section.go#L37)):*
> ```Go
> 
> func (s SectionID) name() string {
> 	n, ok := map[SectionID]string{
> ```

Make it a global map, and method can be named String, so print will render these names

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/go-interpreter/wagon/8)*

dennwc:
Reviewed 3 of 3 files at r11.
Review status: all files reviewed at latest revision, 68 unresolved discussions.

---

*[wasm/section.go, line 621 at r8](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn5TBDnJ5Q9P8THNEp0:-Kn96Q2V89aFgki_IuMp:b-xfludm) ([raw file](https://github.com/go-interpreter/wagon/blob/d51097470837f3534c28ed859303afcaee27a59f/wasm/section.go#L621)):*

`ReadAll` never returns `EOF`
</blockquote></details>

On a second though, `ReadFull` is more appropriate here, since it will ask exactly for `bodySize` bytes, and if less is received it will throw `UnexpectedEOF`.

Also `ReadFull` forces you to allocate exactly the buffer you need, while `ReadAll` grows slice continuously and creates more garbage because of this. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/go-interpreter/wagon/8)*

vibhavp:
Review status: 8 of 10 files reviewed at latest revision, 70 unresolved discussions.

---

*[wasm/imports.go, line 25 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVp-r1-25:-Kn5y1LgAFebLeI5R9ML:bc43lj8) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/imports.go#L25)):*

also: perhaps move this into the `wasm/errors.go` file ?
</blockquote></details>

ah, I was confusing convention for types and vars, done.

---

*[wasm/imports.go, line 58 at r8](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn5B8pmRr07oOu5iGZE:-Kn6-nODH_IQg44A81vu:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/d51097470837f3534c28ed859303afcaee27a59f/wasm/imports.go#L58)):*

Specify field name in error
</blockquote></details>

Done.

---

*[wasm/index.go, line 1 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVq-r1-1:-Kn6-o8HutAz9QCDKtuY:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/index.go#L1)):*

license header
</blockquote></details>

Done.

---

*[wasm/index.go, line 19 at r8](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn5BbncdS0MpOXcoRTP:-Kn87j_djW8J2oP5CRMC:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/d51097470837f3534c28ed859303afcaee27a59f/wasm/index.go#L19)):*

Invert condition and place it as "or" condition to an if above
</blockquote></details>

Done.

---

*[wasm/index.go, line 95 at r8](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn5CElBsqm7GMtVa4EP:-KnDghJxCcrOUjHvkxYU:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/d51097470837f3534c28ed859303afcaee27a59f/wasm/index.go#L95)):*

This looks unclear at a first glance. How about reallocating slice with a new size and `copy` an old one to it?
</blockquote></details>

Done.

---

*[wasm/index.go, line 150 at r8](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn5D4gidYkESinFlByY:-KnDbFQySCir6hx_1PtV:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/d51097470837f3534c28ed859303afcaee27a59f/wasm/index.go#L150)):*

Same as above - allocate with a new size instead of doing index calculations to grow the slice. Later it will make sense to scan all linear memory indexes to get a maximal offset+size, allocate once an copy everything there.
</blockquote></details>

Done.

---

*[wasm/index.go, line 153 at r8](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn5CvR8BJZsv4q1Wd1D:-KnDbFeBc6GDI_dxVmfv:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/d51097470837f3534c28ed859303afcaee27a59f/wasm/index.go#L153)):*

Don't shadow builtins and use `copy` instead of a loop
</blockquote></details>

Done.

---

*[wasm/index.go, line 15 at r10](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn6Axru_I9f_pXYPJAy:-KnDbG7iC1T4DJsSo8CL:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/9b6b3bef1d94ca2b5051b2e768676ec854ca1bc9/wasm/index.go#L15)):*

It's better to include a cast here: `int(e)`. It's dangerously close to infinite recursion (`print` will try to call `e.Error` if format placeholder is different).
</blockquote></details>

Done.

---

*[wasm/init_expr.go, line 1 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVr-r1-1:-Kn6-td_Rp_7hJAwgDfR:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/init_expr.go#L1)):*

license header
</blockquote></details>

Done.

---

*[wasm/init_expr.go, line 24 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVr-r1-24:-Kn6-uWo3I2mbYw293kr:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/init_expr.go#L24)):*

we usually just call `io.Reader` variables `r`.
</blockquote></details>

Done.

---

*[wasm/init_expr.go, line 26 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVr-r1-26:-Kn6-v4A-Slz7BMW-Khv:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/init_expr.go#L26)):*

```go
buf := new(bytes.Buffer)
```

works just as well.
</blockquote></details>

Done.

---

*[wasm/section.go, line 97 at r8](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn5QAJoHhfuNZdJ7tuR:-KnBkwdmgarIcE1CUBss:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/d51097470837f3534c28ed859303afcaee27a59f/wasm/section.go#L97)):*

Place a TODO to read it from stream directly without allocation the whole buffer.
</blockquote></details>

Done.

---

*[wasm/section.go, line 621 at r8](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn5TBDnJ5Q9P8THNEp0:-KnBjtlVmB6_nPjfuG2x:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/d51097470837f3534c28ed859303afcaee27a59f/wasm/section.go#L621)):*

On a second though, `ReadFull` is more appropriate here, since it will ask exactly for `bodySize` bytes, and if less is received it will throw `UnexpectedEOF`.

Also `ReadFull` forces you to allocate exactly the buffer you need, while `ReadAll` grows slice continuously and creates more garbage because of this. 
</blockquote></details>

Done.

---

*[wasm/types.go, line 240 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVw-r1-240:-KnDbAj1wx13ZxcCYKzj:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/types.go#L240)):*

perhaps document a bit more what `ResizableLimits` is?
</blockquote></details>

Done.

---

*[wasm/types.go, line 266 at r8](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn5XDwAcTYgIQaa50Vb:-KnDb9r15LIgQ1pPu5f8:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/d51097470837f3534c28ed859303afcaee27a59f/wasm/types.go#L266)):*

Later it will look like this: `flags & 0x32 == 0x32`. It's better to not duplicate constants in the expression: `flags & 0x32 != 0`. Same in case of `0x1`.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/go-interpreter/wagon/8)*

dennwc:
Reviewed 1 of 1 files at r12, 2 of 2 files at r13.
Review status: all files reviewed at latest revision, 62 unresolved discussions.

---

*[wasm/section.go, line 97 at r8](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn5QAJoHhfuNZdJ7tuR:-KnEGcMMxfPlEMsNMH0h:b-rcveqq) ([raw file](https://github.com/go-interpreter/wagon/blob/d51097470837f3534c28ed859303afcaee27a59f/wasm/section.go#L97)):*
<details><summary><i>Previously, vibhavp (Vibhav Pant) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Great! One last thing here - wrap `r` into `LimitReader` with `payloadDataLen`, and Grow buffer to the same length.

Later we can add a flag that omits reading to buffer, so it will be just `sectionReader := io.LimitReader(r, payloadDataLen)`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/go-interpreter/wagon/8)*

vibhavp:
Review status: all files reviewed at latest revision, 62 unresolved discussions.

---

*[wasm/section.go, line 97 at r8](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn5QAJoHhfuNZdJ7tuR:-KnIENigP4Ej2R8J9VD5:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/d51097470837f3534c28ed859303afcaee27a59f/wasm/section.go#L97)):*

Great! One last thing here - wrap `r` into `LimitReader` with `payloadDataLen`, and Grow buffer to the same length.

Later we can add a flag that omits reading to buffer, so it will be just `sectionReader := io.LimitReader(r, payloadDataLen)`.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/go-interpreter/wagon/8)*

sbinet:
Review status: 9 of 10 files reviewed at latest revision, 13 unresolved discussions.

---

*[wasm/init_expr.go, line 42 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVr-r1-42:-KnIz9FN-evu1LoDMM0r:bwksh7m) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/init_expr.go#L42)):*

Still unresolved. Write `b` directly to buf.
</blockquote></details>

yeah, something like that:
```go
_, err = buf.Write(b)
```

---

*[wasm/init_expr.go, line 49 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVr-r1-49:-KnIzXEO3Jqq4RyxQ6nV:b-ez1c49) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/init_expr.go#L49)):*

I agree here, it's better to use `binary.LittleEndian.PutUint32` directly, it will be much faster.

Also, I still don't understand why you re-encode the expression to a different format to execute it later.  If you want to have it in Buffer, you can make an `io.TeeReader`, and it will automatically populate the buffer while you reading this expression.
</blockquote></details>

one could create a `tmp [8]byte` scratch space one would write to via the `binary.LittleEndian.PutUint{32,64}` and then use an `append` or `buf.Write(tmp[:])`

---

*[wasm/init_expr.go, line 115 at r8](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn5FUjmeLhDCDpBPmlk:-KnJ-6MvnN1d3HOmHIsl:btjx2wq) ([raw file](https://github.com/go-interpreter/wagon/blob/d51097470837f3534c28ed859303afcaee27a59f/wasm/init_expr.go#L115)):*

Same here - avoid `binary.Read`, it uses reflection which is slow.
</blockquote></details>

yeah, like for the `Write` case. or mark it as a TODO/FIXME (with a link to an issue you had filed beforehand so we don't forget)

---

*[wasm/module.go, line 16 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVt-r1-16:-KnJ-OlwOKm7p8Ziy78b:b3gg0d3) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/module.go#L16)):*

please document `Module`.
at the very least you should point to the specifications of WASM for a module.
</blockquote></details>

I would perhaps insert a `\n` after `module.` and then add the link to the wasm modules docs on a new line (so it looks better in godoc)

---

*[wasm/section.go, line 53 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVv-r1-53:-KnJ0QwEV3vJYr9zz7H_:b7ia8be) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/section.go#L53)):*

please document what the boolean means.
</blockquote></details>

using the canonical way to document functions, it would read:

```go
// readSection reads a valid section ...
```

---

*[wasm/section.go, line 211 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVv-r1-211:-KnJ0pzEutwjRRtK4UAP:b8488q0) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/section.go#L211)):*

no need to declare `err` here. it was done at line 204.
</blockquote></details>

still there :)

---

*[wasm/section.go, line 603 at r8](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn5SYnfkGIEiJEfbdzJ:-KnJ6WTC1pqr4gP4xo6F:b-3zue7m) ([raw file](https://github.com/go-interpreter/wagon/blob/d51097470837f3534c28ed859303afcaee27a59f/wasm/section.go#L603)):*

It might be a good idea to `defer io.Copy(ioutil.Discard, r)` to drain remaining body if it had some errors in it. In this case the next section might be read without errors and this function can be marked as damaged, for example.
</blockquote></details>

I would also add that `defer`, if only for debugging purposes (so that errors in the decoding of one section doesn't silently percolate to others)

---

*[wasm/section.go, line 37 at r10](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn6BvrzoGs74qdmI4Ai:-KnJ00QPVAbgNyAWdD_5:bcslvw2) ([raw file](https://github.com/go-interpreter/wagon/blob/9b6b3bef1d94ca2b5051b2e768676ec854ca1bc9/wasm/section.go#L37)):*

Make it a global map, and method can be named String, so print will render these names
</blockquote></details>

or use a switch statement?
or a global slice of strings (as the `SectionID`s values are compact)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/go-interpreter/wagon/8)*

vibhavp:
Review status: 9 of 10 files reviewed at latest revision, 35 unresolved discussions.

---

*[wasm/imports.go, line 22 at r8](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn5AiRDag9PQw1rSRSc:-KnK478MzYRYZ0bLeC2k:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/d51097470837f3534c28ed859303afcaee27a59f/wasm/imports.go#L22)):*

It can be replaced with type enum to force compiler to check type on this field. In this case Kind may also be omitted in favor of Go type system:

```golang
// Import is a type enumeration implemented by:
// - FuncImport
// - TableImport
// - MemoryImport
// - GlobalImport
type Import interface{
  isImport() // ensures that only this package can implement it
}

type FuncImport struct{
  Type uint32
}
func (FuncImport) isImport() {}

type TableImport struct{
  Type Table
}
func (TableImport) isImport() {}

type MemoryImport struct{
  Type Memory
}
func (MemoryImport) isImport() {}

type GlobalImport struct{
  Type GlobalVar
}
func (GlobalImport) isImport() {}
```

This way it will preserve type safety and prevent some potential bugs.
</blockquote></details>

Done.

---

*[wasm/init_expr.go, line 77 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVr-r1-77:-KnK4BmcJnBUnLtvoHGr:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/init_expr.go#L77)):*

`s/opecode/opcode/`
</blockquote></details>

Done.

---

*[wasm/module.go, line 1 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVt-r1-1:-KnL9Zza89XLrp8cy2nw:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/module.go#L1)):*

here would be also a good place (or put it under a `wasm/doc.go` file) for the package-level documentation.
</blockquote></details>

Done.

---

*[wasm/module.go, line 86 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVt-r1-86:-KnL9ePr6uqf8sLTqnCR:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/module.go#L86)):*

perhaps write it like that:

```go
for _, fct := range []func() error{
    m.populateGlobals,
    m.populateFunctions,
    m.populateTables,
    m.populateLinearMemory,
}{
  err := fct()
  if err != nil {
     return nil, err
  }
}
```
</blockquote></details>

Done.

---

*[wasm/section.go, line 1 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVv-r1-1:-KnL9fonNoQmKyxFaj45:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/section.go#L1)):*

please add a license header
</blockquote></details>

Done.

---

*[wasm/section.go, line 45 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVv-r1-45:-KnL9gUAil8xe8N1MtTY:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/section.go#L45)):*

`s/ErrInvalidSectionID/InvalidSectionIDError/`

also, perhaps move it to `wasm/errors.go`
</blockquote></details>

Done.

---

*[wasm/section.go, line 53 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVv-r1-53:-KnLA7dwSU2IAjEboBua:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/section.go#L53)):*

using the canonical way to document functions, it would read:

```go
// readSection reads a valid section ...
```
</blockquote></details>

Done.

---

*[wasm/section.go, line 64 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVv-r1-64:-KnL9gubcpM5sO2w9DK9:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/section.go#L64)):*

I would re-write this as:
```go
s := Section{ID: SectionID(id)}
```
and remove the earlier declaration.
</blockquote></details>

Done.

---

*[wasm/section.go, line 210 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVv-r1-210:-KnLA-DGtoXqkD5SA07c:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/section.go#L210)):*

ditto:
```go
for i := range s.Entries { ... }
```
</blockquote></details>

Done.

---

*[wasm/section.go, line 211 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVv-r1-211:-KnLA-fS6C5txrtl107x:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/section.go#L211)):*

still there :)
</blockquote></details>

Done.

---

*[wasm/section.go, line 286 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVv-r1-286:-KnLA0b_Dkb11pcwNqE7:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/section.go#L286)):*

I would use a `for-range` on `s.Types` here (and create the slice with the correct length, like for the other `readSectionXXX`)
</blockquote></details>

Done.

---

*[wasm/section.go, line 312 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVv-r1-312:-KnLA0tYD4_By8UgTht-:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/section.go#L312)):*

I would use a `for-range` on `s.Entries`.
</blockquote></details>

Done.

---

*[wasm/section.go, line 339 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVv-r1-339:-KnLA1J0_qkoad3yT_Yb:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/section.go#L339)):*

use a `for-range` on `s.Entries`.
</blockquote></details>

Done.

---

*[wasm/section.go, line 367 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVv-r1-367:-KnLA1aJIOjPeDEmuilE:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/section.go#L367)):*

use a `for-range` on `s.Globals`.
</blockquote></details>

Done.

---

*[wasm/section.go, line 380 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVv-r1-380:-KnLA21T02zx0Ddsfgbw:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/section.go#L380)):*

```go
Type *GlobalVar // Type holds informations about the type and the mutability of this global variable
```
</blockquote></details>

Done.

---

*[wasm/section.go, line 456 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVv-r1-456:-KnLA3JhDGTHbu_RUdwq:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/section.go#L456)):*

please document the index into which section.
</blockquote></details>

Done.

---

*[wasm/section.go, line 486 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVv-r1-486:-KnLA3aPUfCwhW543Mgg:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/section.go#L486)):*

please use a `for-range` into `s.Entries`.
</blockquote></details>

Done.

---

*[wasm/section.go, line 497 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVv-r1-497:-KnLA3y3z14Zzoxp9Gt_:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/section.go#L497)):*

please document `ElementSegment`.
</blockquote></details>

Done.

---

*[wasm/section.go, line 521 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVv-r1-521:-KnLA4dQ_24vMb3dQ3oz:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/section.go#L521)):*

please use a `for-range` into `s.Elems`.
</blockquote></details>

Done.

---

*[wasm/section.go, line 548 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVv-r1-548:-KnLA5y-09_YiuMpGFt_:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/section.go#L548)):*

please use a `for-range` into `s.Bodies`.
</blockquote></details>

Done.

---

*[wasm/section.go, line 602 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVv-r1-602:-KnLA6F1GE7BhTJk0axG:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/section.go#L602)):*

please use a `for-range` into `f.Locals`.
</blockquote></details>

Done.

---

*[wasm/section.go, line 661 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVv-r1-661:-KnLA7-TdpopRj7dDTF3:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/section.go#L661)):*

please use a `for-range` into `s.Entries`.
</blockquote></details>

Done.

---

*[wasm/section.go, line 603 at r8](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn5SYnfkGIEiJEfbdzJ:-KnLA9Eo5aYvscOdO3fp:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/d51097470837f3534c28ed859303afcaee27a59f/wasm/section.go#L603)):*

I would also add that `defer`, if only for debugging purposes (so that errors in the decoding of one section doesn't silently percolate to others)
</blockquote></details>

Done.

---

*[wasm/types.go, line 123 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVw-r1-123:-KnLAgh0J5aBPvbR-ZxW:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/types.go#L123)):*

please use a `for-range` into `f.ParamTypes`.
</blockquote></details>

Done.

---

*[wasm/types.go, line 135 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVw-r1-135:-KnLAh6NT79-QHKmhu7w:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/types.go#L135)):*

please use a `for-range` into `f.ReturnTypes` (and create it with the expected `returnCount` length)
</blockquote></details>

Done.

---

*[wasm/errors.go, line 1 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVo-r1-1:-KnLAi0Uo_MKB1siek8g:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/errors.go#L1)):*

please add a license header.
</blockquote></details>

Done.

---

*[wasm/errors.go, line 7 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVo-r1-7:-KnLAigAfuFUwEjyAi_p:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/errors.go#L7)):*

please add some documentation about this type and its fields.
looking at the implementation of its `Error() string` method, one can infer the meaning of these fields, but one shouldn't have to look under the hood :)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/go-interpreter/wagon/8)*

vibhavp:
Review status: 8 of 10 files reviewed at latest revision, 10 unresolved discussions.

---

*[wasm/section.go, line 184 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVv-r1-184:-KnNmaglBhVHUNt9icuQ:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/section.go#L184)):*

I would write it:
```go
for i := range s.Entries {
   // ...
```

instead, to help the compiler to remove the index check.
</blockquote></details>

Done.

---

*[wasm/section.go, line 268 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVv-r1-268:-KnNmbUpB8T34smpQKyx:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/section.go#L268)):*

I would perhaps add an `err != nil` check here to prevent any accidental clobbering of that variable.
or, alternatively, add the `err` checks in each `case` ?
</blockquote></details>

Done.

---

*[wasm/section.go, line 37 at r10](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn6BvrzoGs74qdmI4Ai:-KnNmWkLxWmEMx8GGJQi:bnco3yu) ([raw file](https://github.com/go-interpreter/wagon/blob/9b6b3bef1d94ca2b5051b2e768676ec854ca1bc9/wasm/section.go#L37)):*

or use a switch statement?
or a global slice of strings (as the `SectionID`s values are compact)
</blockquote></details>

Done. Renamed name() to String()

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/go-interpreter/wagon/8)*

vibhavp:
Review status: 8 of 10 files reviewed at latest revision, 10 unresolved discussions.

---

*[wasm/init_expr.go, line 49 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVr-r1-49:-KnNnsqiG3o9oxx5EE_0:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/init_expr.go#L49)):*

one could create a `tmp [8]byte` scratch space one would write to via the `binary.LittleEndian.PutUint{32,64}` and then use an `append` or `buf.Write(tmp[:])`
</blockquote></details>

Done.

---

*[wasm/init_expr.go, line 94 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVr-r1-94:-KnNo-Q7VtrD5Q0r5uVB:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/init_expr.go#L94)):*

just call it `r`.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/go-interpreter/wagon/8)*

sbinet:
Reviewed 1 of 9 files at r2, 2 of 8 files at r6, 1 of 3 files at r11, 2 of 3 files at r16.
Review status: 8 of 10 files reviewed at latest revision, 9 unresolved discussions.

---

*[wasm/init_expr.go, line 178 at r16](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-KnOJOVSSOnxEA97EBx4:-KnOJOVSSOnxEA97EBx5:b-f1beuy) ([raw file](https://github.com/go-interpreter/wagon/blob/43c7f1441a611b99d197bd2829195c0dbe79e8c5/wasm/init_expr.go#L178)):*
> ```Go
> 		return math.Float64frombits(uint64(v)), nil
> 	default:
> 		panic("Invalid lastVal value")
> ```

perhaps display that invalid value?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/go-interpreter/wagon/8)*

vibhavp:
Review status: 8 of 10 files reviewed at latest revision, 9 unresolved discussions.

---

*[wasm/init_expr.go, line 178 at r16](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-KnOJOVSSOnxEA97EBx4:-KnOUrO8fMCAxFj10vrN:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/43c7f1441a611b99d197bd2829195c0dbe79e8c5/wasm/init_expr.go#L178)):*

perhaps display that invalid value?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/go-interpreter/wagon/8)*

dennwc:
One small change left in `execInitExpr`, and I think we are good to merge.

---

Reviewed 1 of 1 files at r14, 1 of 3 files at r16, 1 of 1 files at r17.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-:-KnOcy2fuEiuXLZRQGGL:b-2ctrbk)*

vibhavp:
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[wasm/init_expr.go, line 107 at r8](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn5FH2No8zdvvrpHoE6:-KnOqE0QYkcs2olAWJq_:b-896fix) ([raw file](https://github.com/go-interpreter/wagon/blob/d51097470837f3534c28ed859303afcaee27a59f/wasm/init_expr.go#L107)):*

```golang
if err == io.EOF {
  break
} else if err != nil {
  return nil, err
}
```
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/go-interpreter/wagon/8)*

dennwc:
:lgtm:

---

Reviewed 1 of 1 files at r18.
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-:-KnQrI-8Fm61AQZ0IVeK:bnfp4nl)*

sbinet:
:lgtm:

---

Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[wasm/init_expr.go, line 49 at r1](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn4G1U_jCiABHKxqqVr-r1-49:-KnQtS74AnGKwKVIcNyD:b-3e70zd) ([raw file](https://github.com/go-interpreter/wagon/blob/0fd988865ca656bb6802542d094a408a398f4170/wasm/init_expr.go#L49)):*
<details><summary><i>Previously, vibhavp (Vibhav Pant) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

let's do the optimization (bypassing `binary.Read`) in another PR.

---

*[wasm/init_expr.go, line 115 at r8](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-Kn5FUjmeLhDCDpBPmlk:-KnQtlcRhgoSI2jiHGmO:bxrn29z) ([raw file](https://github.com/go-interpreter/wagon/blob/d51097470837f3534c28ed859303afcaee27a59f/wasm/init_expr.go#L115)):*

yeah, like for the `Write` case. or mark it as a TODO/FIXME (with a link to an issue you had filed beforehand so we don't forget)
</blockquote></details>

(like for `readInitExpr`, let's leave the optimization for another PR)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/go-interpreter/wagon/8#-:-KnQuOgD_H16byybXUBr:bnfp4nl)*

shurcooL:
Just wondering, but why not simply merge this PR via squash strategy? That way, it'll be merged to `master` branch as a single commit, but all the commits and the associated review comments in this PR will be preserved.

sbinet:
@shurcooL this makes for a much nicer, linear git history.

PR #17, opened 2017-08-08
validate,wasm: remove use of slow binary.Read

sbinet:
This CL removes the calls to binary.Read(r, bo, &v) where v is a simple
numeric type.
It replaces it with explicit calls to binary.Order.Uint{32|64} that
don't involve calls to reflect.

---
This change is [Reviewable](https://reviewable.io/reviews/go-interpreter/wagon/17)


vibhavp:
:lgtm:

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/go-interpreter/wagon/17#-:-Kr15-pRa-rhvKuOMdnE:bnfp4nl)*

wasm/read.go
```diff
@@ -25,3 +26,21 @@ func readString(r io.Reader, n int) (string, error) {
...
+	var buf [4]byte
+	_, err := io.ReadFull(r, buf[:])
+	if err != nil {
+		return 0, err
+	}
+	return binary.LittleEndian.Uint32(buf[:]), nil
+}
+
+func readU64(r io.Reader) (uint64, error) {
+	var buf [4]byte
```
shurcooL: Shouldn't this be `var buf [8]byte`? `uint64` is 8 bytes.

sbinet: yes. good catch. copy-paste hit me yet again :)

shurcooL: Followup question, how come tests didn't catch this? I'd expect things to fail badly as a result of a low level primitive like this being incorrect. Is this code not used/tested significantly yet? (That's completely fine if so, I just want to learn.)

sbinet: addressed w/ go-interpreter/wagon#18

sbinet: b/c there isn't any *yet*. at least not for `wasm` itself.
(I was in the process of adding some. see go-interpreter/wagon#19)

shurcooL: Makes sense. Thanks!

PR #19, opened 2017-08-08
wasm/internal/readpos: add a test for ReadPos.Read

sbinet:
This change is [Reviewable](https://reviewable.io/reviews/go-interpreter/wagon/19)


.travis.yml
```diff
@@ -1,8 +1,11 @@
 language: go
 go:
-  - 1.6.x
   - 1.7.x
   - 1.8.x
   - master
 
+matrix:
+ allow_failures:
+   - go: master
```
shurcooL: You might want to enable `fast_finish` here as well, so build status is reported more quickly.

See https://github.com/shurcooL/home/blob/e6f2e71819d2db3cd6fb1a781ac880cbfff3ed9b/.travis.yml#L9 for example.

See https://blog.travis-ci.com/2013-11-27-fast-finishing-builds/ for reference.

sbinet: thanks for the tip.
done. :)

PR #21, opened 2017-08-09
wagon: add go-vet test

sbinet:
This change is [Reviewable](https://reviewable.io/reviews/go-interpreter/wagon/21)


shurcooL:
This is interesting. I've never seen `go vet` and `gofmt` checks implemented as Go tests.

Commonly, they're implemented as commands that run in CI. For example, see [here](https://github.com/gorilla/mux/blob/ac112f7d75a0714af1bd86ab17749b31f7809640/.travis.yml#L20-L21).

What's the reason you've chosen to implement it this way?

sbinet:
manyfold:

- one fewer command to remember
- it's part of `go test`
- the failure of the test (if any) is closer to the developer
- travis-ci is less of a snowflake
- it's Go code and not some incantation lost somewhere in some travis-only configuration file

PR #22, opened 2017-08-10
wasm: add first batch of tests

sbinet:
This CL adds a test for ReadModule and pairs for .wasm/.wast files.
The .wasm files have been generated with 'wast2wasm' from
the github.com/WebAssembly/wabt project.

---
This change is [Reviewable](https://reviewable.io/reviews/go-interpreter/wagon/22)


sbinet:
@shurcooL PTAL

wasm/module_test.go
```diff
@@ -0,0 +1,34 @@
...
+	"bytes"
+	"io/ioutil"
+	"path/filepath"
+	"testing"
+
+	"github.com/go-interpreter/wagon/wasm"
+)
+
+func TestReadModule(t *testing.T) {
+	fnames, err := filepath.Glob("testdata/*wasm")
```
shurcooL: Maybe `filepath.Glob("testdata/*.wasm")` to make it more clear you're looking for all files with `.wasm` extension.

Also, I'm not sure how `filepath.Glob` operates on Windows and whether it'll work with forward slash separators. But I'm okay if you don't think it's worth dealing with at this stage.

wasm/module_test.go
```diff
@@ -0,0 +1,34 @@
...
+		raw, err := ioutil.ReadFile(fname)
+		if err != nil {
+			t.Error(err)
+			continue
+		}
+		r := bytes.NewReader(raw)
+		_, err = wasm.ReadModule(r, nil)
+		if err != nil {
+			t.Errorf("error reading module %q: %v", fname, err)
+			continue
```
shurcooL: This `continue` isn't neccessary since you're not doing anything after. But it would be, if you were.

I'd avoid writing it until it's actually needed because it's a no-op. Either way is fine.

wasm/module_test.go
```diff
@@ -0,0 +1,34 @@
...
+		}
```
shurcooL: If you wanted to, you could modify line above to `m, err = wasm.ReadModule(r, nil)`, and add a quick test here that `m != nil`:

```Go
if m == nil {
	t.Errorf("wasm.ReadModule of %q return nil module and nil error", fname)
}
```

(Then, the `continue` above would be needed.)

wasm/module_test.go
sbinet: done.

wasm/module_test.go
sbinet: done.

wasm/module_test.go
sbinet: done.

sbinet:
PTAL

wasm/module_test.go
```diff
@@ -14,21 +14,29 @@ import (
...
-			t.Error(err)
-			continue
-		}
-		r := bytes.NewReader(raw)
-		_, err = wasm.ReadModule(r, nil)
-		if err != nil {
-			t.Errorf("error reading module %q: %v", fname, err)
-			continue
-		}
+		testModule(t, fname)
```
shurcooL: What's the reason you chose to to factor all this out into `testModule`? It doesn't seem to be helpful. You're not making use of any `defer`s. All `return`s can just be replaced with `continue`s and it'd be just fine inside the `TestReadModule` for loop. I don't think it helps, it only hurts readability IMO. But it's not a huge deal.

sbinet: I thought that with refactoring into `testModule`, the `return`s would be more natural-looking Go code.

I put a `defer` for good measure :)

wasm/module_test.go
```diff
@@ -0,0 +1,42 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	for _, fname := range fnames {
+		testModule(t, fname)
+	}
+}
+
+func testModule(t *testing.T, fname string) {
+	f, err := os.Open(fname)
```
shurcooL: Have you considered the impact this change makes on the test behavior?

If there is an error reading the file half-way through, previously, it would've been reported as an error reading the file, and `wasm.ReadModule` would not even be called.

But now, the error will come from `wasm.ReadModule`, since the `io.Reader` given to it may fail.

Both ways have tradeoffs. I just wanted to make sure this was an intentional change.

(P.S. I was leaving this comment on the 3rd commit, but you've squashed & force pushed while I was writing it, so I have to post it on the final diff.)

sbinet: (sorry about that.)

sbinet: you're right.

mid-flight, I remembered I had dropped go1.6 from the test-matrix (as Google AppEngine will *soon* (right?) migrate to go1.8, once go1.9 is out).
so I can use sub-tests.

shurcooL: 👍 to subtests and bringing the actual test code back into `TestReadModule`. It's more readable now.

PR #43, opened 2017-12-12
Access VM memory

### LINKED ISSUE #42: VM: allow access to linear memory
Currently it looks like we can get public access to a `Module` linear memory: https://github.com/go-interpreter/wagon/blob/master/wasm/module.go#L50 which is a static view of the memory space after parsing and loading the module. However it doesn't appear that VM linear memory is copied back into the module after functions are executed.

It would be very useful to be able to access the memory of a `VM` between execution of functions: https://github.com/go-interpreter/wagon/blob/master/exec/vm.go#L59. This would allow interoperability between a running module and an embedding Go program, for example.

Barring any objections I will probably make an attempt to implement a function that exposes the linear memory of a `VM` instance.

bkeroackdsc:
Allow access to VM memory. Fixes #42

---
This change is [Reviewable](https://reviewable.io/reviews/go-interpreter/wagon/43)


bkeroackdsc:
Rebased and re-formatted commit message per request in #45

Issue #40, opened 2017-12-05
wasm: implement imports


sbinet:
`wasm.ReadModule` needs to be able to handle `imports`.

perhaps `wasm.ResolveFunc` should be an interface instead of a `type X func` to be able to import builtin functions (_e.g._ `print`)

sbinet:
@vibhavp do you want (and/or have the cycles) to have a stab at it?
(just asking to reduce the amount of duplicated work)

vibhavp:
If you take a look at the `wasm` fork, it uses JavaScript to provide access to some runtime functions that cannot be implemented alone in WebAssembly. We could allow the table of a `wasm.Module` to be populated with Go functions. Similarly, `exec` would also need some modification so as to call Table elements with `reflect`. 

Making `ResolveFunc`  an interface would only allow for a fixed number of functions to be imported, while the WebAssembly spec also allows importing variables, tables and memory elements.

sbinet:
ping @vibhavp ?

Robert's work (bringing `GOARCH=wasm` to the official gc toolchain) is starting to percolate.
it would be great to have this working for Go-1.11.x :)

PR #44, opened 2017-12-13
wasm: implement imports

sbinet:
Updates go-interpreter/wagon#40.

---
This change is [Reviewable](https://reviewable.io/reviews/go-interpreter/wagon/44)


vibhavp:
:lgtm:

In the future, we could allow `wasm.Module` to have Go-native functions and variables.

---

Review status: 0 of 5 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/go-interpreter/wagon/44#-:-L0K78wQEahtl2Gfr2g_:bcq5hl6)*

sbinet:
yep, that would be great :)
(so the last step of my advent-2017 blog post could be done in pure Go instead of a web server executing that `wasm` module or reaching for `nodejs` to do the same...)

PR #53, opened 2018-03-01
Return an error on Unreachable instead of panicing

gballet:
So far, when the vm finds an unreachable instruction, it panics. This is a problem for whoever wants to embed wagon inside their binaries, which is exactly what the Ethereum eWASM team considers at the moment.

As a result, this PR adds an `error` return to each opcode function. Those opcode errors are then returned to the client code so that the caller can decide on the best course of action.


sbinet:
I am not against this change _per se_ (I also don't like when libraries panic) but I am curious about the use case.
Could you expand a bit?

(also, please adapt the `TestSpec` in `exec/exec_test.go` for the new semantics introduced with returning an error instead of panicking)

gballet:
>  I am curious about the use case.

I'm embedding wagon into the Ethereum go client as part of a research project inside the Ethereum Foundation to replace the ethereum virtual machine with wasm. I just use wagon as a library, I set up a vm, load my module and finally call `vm.ExecCode`.
The problem with a `panic` is that if it's placed inside the "library" part of the dependency, then I inherit it, which is going to take the whole system down. I want to be able to run an invalid module, detect an error and handle it in a safe way. Hence this change.

gballet:
@sbinet I am not able to reproduce the problem in Travis locally running `go test -v ./...`. Is there a special env setting that I should be aware of ?

sbinet:
cherry-picking your change on top of the latest `master`:

```
$> git l
* c75847b (HEAD -> no-panics) Return an error on Unreachable instead of panicing
* 6a484e5 (origin/master, master) wasm: make logging non-verbose by default
* 6947051 travis: add 1.10.x
* 0fab5d2 cmd/wasm-dump, wasm: do not require an importer
* 6c2b2ee cmd/wasm-{dump,run}: implement a naive module resolver
* adcf76a wasm: implement imports
* 3d087b0 wasm: fix infinite recursion in InvalidFunctionIndexError.Error
* 1698ff5 exec: export the VM memory field
```

I get:
```
$> cd $GOPATH/src/github.com/go-interpreter/wagon
$> go test ./...
ok  	github.com/go-interpreter/wagon	1.307s
?   	github.com/go-interpreter/wagon/cmd/wasm-dump	[no test files]
?   	github.com/go-interpreter/wagon/cmd/wasm-run	[no test files]
?   	github.com/go-interpreter/wagon/cmd/wasm-srv	[no test files]
?   	github.com/go-interpreter/wagon/cmd/wasm-test-srv	[no test files]
?   	github.com/go-interpreter/wagon/disasm	[no test files]
--- FAIL: TestSpec (0.01s)
    --- FAIL: TestSpec/testdata/spec/select.wasm (0.01s)
    	exec_test.go:290: /home/binet/work/gsoc/src/github.com/go-interpreter/wagon/exec/testdata/spec/select.wasm, select_trap_l: exec: reached unreachable
    --- FAIL: TestSpec/testdata/spec/unwind.wasm (0.02s)
    	exec_test.go:290: /home/binet/work/gsoc/src/github.com/go-interpreter/wagon/exec/testdata/spec/unwind.wasm, func-unwind-by-unreachable: exec: reached unreachable
FAIL
FAIL	github.com/go-interpreter/wagon/exec	0.540s
?   	github.com/go-interpreter/wagon/exec/internal/compile	[no test files]
?   	github.com/go-interpreter/wagon/internal/stack	[no test files]
?   	github.com/go-interpreter/wagon/validate	[no test files]
ok  	github.com/go-interpreter/wagon/wasm	0.008s
ok  	github.com/go-interpreter/wagon/wasm/internal/readpos	0.008s
ok  	github.com/go-interpreter/wagon/wasm/leb128	0.003s
ok  	github.com/go-interpreter/wagon/wasm/operators	0.002s
```

from which directory are you executing your `go test ./...` ?

sbinet:
as for the use cases, thanks.

I was wondering whether one couldn't be sure, once a module was validated, that it wouldn't panic.
or wrap the vm.ExecCode inside a `defer+recover`...

but biting the bullet and having the ops return an error is, I guess, fine.
@vibhavp, @shurcooL  do you agree?

gballet:
Alright, indeed `master` had drifter away from my branch and I didn't notice. I'm going to fix the tests, rebase and update the pull request. Thanks.

gballet:
> I was wondering whether one couldn't be sure, once a module was validated, that it wouldn't panic.
or wrap the vm.ExecCode inside a defer+recover...

I would agree that if the validation function is good, then this shouldn't happen. From my point of view, though, we are going to deal with some code that is potentially malicious and can never be deleted. Reporting the error is the right course of action over trusting validation that can always be tricked by a crafty enough attacker.

Indeed the client code could use `recover`, however that means the client code need to be aware that the call to `vm.ExecCode` might panic and make its own provisions. Returning an error code is more obvious imho.

vibhavp:
As Sebastian said, there's nothing wrong with returning errors instead of panicking per se, but it would also potentially deviate from the WebAssembly standard.

A trap is reported to the outside environment as a _abnormal termination_, so it makes sense to use `panic` as Go doesn't have any other exception like mechanism. Panicking on reaching `unreachable` is also consistent with the standard Go execution that would result in a `RuntimeError` panic, like:

- [division by zero](https://github.com/go-interpreter/wagon/blob/master/exec/testdata/spec/modules.json#L2722)
- [out of bounds memory access](https://github.com/go-interpreter/wagon/blob/6a484e5c2d7a6e31d9fb72fd712ce0cb90e4c3b5/exec/memory.go#L35)
- [invalid or mismatching `call_indirect` operation](https://github.com/go-interpreter/wagon/blob/master/exec/call.go#L60)

shurcooL:
@sbinet Did you mean to ping me or someone else?

I wasn't very familiar with the context of this PR. After having looked, I'm also on the fence. I can see both sides to it.

I'd be more tempted to side closer with the WebAssembly standard rather than deviating from it. But I'm not sure how much of it is up to interpretation.

I will just point out that some funcs in stdlib do panic on invalid input and that's completely acceptable. They just need to document it. For example:

- [`io.CopyBuffer`](https://godoc.org/io#CopyBuffer) - "if it has zero length, CopyBuffer panics."
- [`rand.Intn`](https://godoc.org/math/rand#Intn) - "It panics if n <= 0."
- [`http.ServeMux.Handle`](https://godoc.org/net/http#ServeMux.Handle) - "If a handler already exists for pattern, Handle panics."

(Of course, all those are invalid inputs that are easily avoidable by the user. Not sure if the same applies here.)

So it kinda depends on whether you want to consider something that would panic an "acceptable" input or not. If so, returning error is more appropriate. If not, then documenting that and panicking is fine, and users that need to handle it can do defer+recover.

gballet:
@vibhavp @shurcooL When a fault occur and an error is returned from `vm.ExecCode`, the execution of the wasm code is interrupted and an error is returned to the client code. That client code is free to panic, and I'm more than happy to make sure `wasm-run` does panic as part of this PR.

After thinking about it a while, I think that I get your point of view: if Go code gets translated into wasm, then it would make sense to behave like proper Go code, following a "bug-for-bug compatibility" principle. Let me point out however, that the wasm code could have been generated from a language following a different, safer philosophy.

If this is too much of a head-scratcher, I suggest that the default behavior would be to panic, and I could add a config flag to return an error instead?

gballet:
alright so now it panics again, but in a different location. Hopefully this will address your concerns.

gballet:
@sbinet @vibhavp @shurcooL I would really appreciate an update on my changes, [our project](https://github.com/ewasm/design) depends on this.

shurcooL:
I don't have any further feedback beyond what I said above. I defer to the project owner on this.

gballet:
@vibhavp https://github.com/go-interpreter/wagon/pull/57 corresponds to your suggestion

PR #57, opened 2018-04-19
client code can request the VM to return an error instead of panicking

gballet:
Provides an alternative to https://github.com/go-interpreter/wagon/pull/53 as suggested by @vibhavp

---
This change is [Reviewable](https://reviewable.io/reviews/go-interpreter/wagon/57)


exec/vm.go
```diff
@@ -238,7 +241,20 @@ func (vm *VM) pushFloat32(f float32) {
...
 // fnIndex should be a valid index into the function index space of
 // the VM's module.
-func (vm *VM) ExecCode(fnIndex int64, args ...uint64) (interface{}, error) {
+func (vm *VM) ExecCode(fnIndex int64, args ...uint64) (rtrn interface{}, err error) {
+	// If used as a library, client code should set vm.NoPanic to true
+	// in order to have an error returned.
+	if vm.NoPanic {
+		defer func() {
+			if r := recover(); r != nil {
+				if reflect.TypeOf(r).Name() == "string" {
```
sbinet: that would be better written as:
```go
if r := recover(); r != nil {
	switch e := r.(type) {
	case error:
		err = e
	case string:
		err = errors.New(e)
	}
}
```

exec/vm.go
```diff
@@ -60,6 +61,8 @@ type VM struct {
 	compiledFuncs []compiledFunction
 
 	funcTable [256]func()
+
+	NoPanic bool
```
sbinet: booleans with negation in their name are always a bit harder to grasp.

what about `Recover` or `EnableRecover`?

Also, as it's exported, please document it.

exec/vm.go
shurcooL: What if it's neither of those two types? Then `err` would be left `nil`.

sbinet: right.
then:
```go
if r := recover(); r != nil {
	switch e := r.(type) {
	case error:
		err = e
	default:
		err = fmt.Errorf("%v", e)
	}
}
```

vibhavp:
LGTM

exec/vm.go
```diff
@@ -60,6 +60,8 @@ type VM struct {
 	compiledFuncs []compiledFunction
 
 	funcTable [256]func()
+
+	recoverExecutionPanic bool
```
sbinet: Apologies but I am a bit confused now: I don't see how this could be set to `true` from outside the `exec` package.

exec/vm.go
```diff
@@ -238,7 +240,21 @@ func (vm *VM) pushFloat32(f float32) {
 // ExecCode calls the function with the given index and arguments.
 // fnIndex should be a valid index into the function index space of
 // the VM's module.
-func (vm *VM) ExecCode(fnIndex int64, args ...uint64) (interface{}, error) {
+func (vm *VM) ExecCode(fnIndex int64, args ...uint64) (rtrn interface{}, err error) {
+	// If used as a library, client code should set vm.NoPanic to true
```
sbinet: the `vm.NoPanic` comment does not match the code now :)

exec/vm.go
gballet: indeed, that's what you get for trying to optimize too early :flushed: . Should be fixed now.

exec/vm.go
```diff
@@ -60,6 +60,8 @@ type VM struct {
 	compiledFuncs []compiledFunction
 
 	funcTable [256]func()
+
+	RecoverExecutionPanic bool // Set to `true` to get an error instead of panicing when execution fails, e.g. because of reaching `unreachable`
```
shurcooL: Optional suggestion for how this can be phrased differently:

```Go
// RecoverExecutionPanic controls whether ExecCode method
// recovers execution panics and returns them as errors instead.
// A panic can only occur when executing an invalid VM.
RecoverExecutionPanic bool
```

I'm not sure if the last sentence is factually correct, so please update accordingly, but I think it's a good idea to explain the situations when panics would occur (i.e., it's not meant to catch bugs in the interpreter implementation).

exec/vm.go
```diff
@@ -69,7 +71,7 @@ var endianess = binary.LittleEndian
 
 // NewVM creates a new VM from a given module. If the module defines a
 // start function, it will be executed.
-func NewVM(module *wasm.Module) (*VM, error) {
+func NewVM(module *wasm.Module, recoverExecutionPanic bool) (*VM, error) {
```
shurcooL: I suspect you shouldn't add the `recoverExecutionPanic` field. Most users wouldn't need to set it, so having it be mandatory here would just make the API harder to use. It's not documented here, so users won't know what it does.

Instead, users who want to have `ExecCode` recover from panics can do:

```Go
vm, err := wagon.NewVM(m)
if err != nil {
    // handle error
}
vm.RecoverExecutionPanic = true
```

gballet: Alright, I'll remove the option. variadic options are nice and cool, however I feel introducing them should be its own PR.

exec/vm.go
gballet: Will do. For reference, `panic`s do occur when the VM is valid and execute invalid software (e.g. `unreachable` is reached, which can happen as the instruction could be found in the binary flow of some WASM file)

exec/vm.go
```diff
@@ -84,6 +91,7 @@ func NewVM(module *wasm.Module) (*VM, error) {
 	vm.globals = make([]uint64, len(module.GlobalIndexSpace))
 	vm.newFuncTable()
 	vm.module = module
+	vm.RecoverExecutionPanic = false
```
shurcooL: There isn't a need to do this, since `false` is already the zero value for `bool` type.

exec/vm.go
```diff
@@ -60,6 +60,13 @@ type VM struct {
 	compiledFuncs []compiledFunction
 
 	funcTable [256]func()
+
+	// RecoverExecutionPanic controls whether the `ExecCode` method
+	// recovers from execution panics and returns them as errors
```
shurcooL: The name and comment mention "execution panic", but doesn't really define it.

What is an "execution panic" and how does it differ from a normal Go run-time panic (as defined at https://golang.org/ref/spec#Handling_panics and https://golang.org/ref/spec#Run_time_panics)?

If by "execution panic" you simply refer to normal Go panics that happen while executing a VM, then I would suggest getting rid of "execution panic" and make it just:

```Go
// RecoverPanic controls whether the `ExecCode` method recovers from panics
// during execution of the VM and returns them as errors instead.
// A panic can occur either when executing an invalid VM
// or encountering an invalid instruction, e.g. `unreachable`.
RecoverPanic bool
```

I think that would be more clear and understandable for readers of the documentation.

If "execution panic" refers to something else defined somewhere else, then it'd be great to reference it, because you can see how it might be misinterpreted otherwise.

exec/vm.go
gballet: I'm aware. I always do this so that the code is more understandable to people who get started with a complex codebase. This way, a simple `grep -r RecoverExecutionPanic` will reveal where it is initialized.

exec/vm.go
```diff
@@ -60,6 +60,13 @@ type VM struct {
 	compiledFuncs []compiledFunction
 
 	funcTable [256]func()
+
+	// RecoverPanic controls whether the `ExecCode` method
+	// recovers from a panic and returns it as an error
+	// instead.
+	// An panic can occur either when executing an invalid VM
```
sbinet: `s/An panic/A panic/`

exec/vm.go
```diff
@@ -238,7 +245,21 @@ func (vm *VM) pushFloat32(f float32) {
...
+	// If used as a library, client code should set vm.RecoverPanic to true
+	// in order to have an error returned.
+	if vm.RecoverPanic {
+		defer func() {
+			if r := recover(); r != nil {
+				switch e := r.(type) {
+				case error:
+					err = e
+				default:
+					err = fmt.Errorf("%v", e)
```
sbinet: `s/"%v"/"exec: %v"/`

sbinet:
ah! and for authorship and all that, please send a PR against [go-interpreter/license](https://github.com/go-interpreter/license) adding yourself to `AUTHORS` and `CONTRIBUTORS`. (then I'll merge this one)

sbinet:
Reviewed 1 of 3 files at r2, 1 of 4 files at r3, 1 of 3 files at r5, 2 of 3 files at r6.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io/reviews/go-interpreter/wagon/57)*

gballet:
Updated the comments, squashed and license PR created at https://github.com/go-interpreter/license/pull/5

sbinet:
one thing I hadn't noticed before: why do we have 2 new `unreachable.wasm` files?
- `wasm/testdata/unreachable.wasm`
- `exec/testdata/spec/unreachable.wasm`

they seem to be completely binary equivalent :)
I guess we can drop the `wasm/testdata/unreachable.wasm` one. right ?

gballet:
just removed it + sqashed again

sbinet:
merged as 6fb5730

PR #58, opened 2018-05-08
RFC: Modules containing native functions

gballet:
In its current state, this PR is a request for comments.

The context of this PR:
  * Use the wagon interpreter as a library in a bigger project and use it as a WASM interpreter;
  * The client code wants to make some of its internal functions available to the client.

The approach that is undertaken in this PR is to add a `NativeFunctions` section to `Module`, that lists the "native" functions to be called by the interpreter. It also adds a `Native` flag to the`Function` struct, which is false by default.
When the interpreter encounters a `call` opcode, the `Native` flag is checked. If it is true, then a function is called.

This approach has issues I'm already aware of:
  * `NativeFunctions` is not specified in the standard
  * it makes module version management more difficult

In its current state, the PR has shortcomings that can be fixed:
  * native code might panic
  * it doesn't support calling functions with parameters
  * `call_indirect` is not currently supported

I'm looking forward to hearing your thoughts on the matter.

---
This change is [Reviewable](https://reviewable.io/reviews/go-interpreter/wagon/58)

### DIFF
--- a/exec/call_test.go
+++ b/exec/call_test.go
@@ -1,0 +1,155 @@
+// Copyright 2018 The go-interpreter Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package exec
+
+import (
+	"bytes"
+	"reflect"
+	"testing"
+
+	"github.com/go-interpreter/wagon/wasm"
+)
+
+func TestHostCall(t *testing.T) {
+	const secretValue = 0xdeadbeef
+
+	var secretVariable int = 0
+
+	// a host function that can be called by WASM code.
+	testHostFunction := func() {
+		secretVariable = secretValue
+	}
+
+	m := wasm.NewModule()
+	m.Start = &wasm.SectionStartFunction{Index: 0}
+
+	// A function signature. Both the host and WASM function
+	// have the same signature.
+	fsig := wasm.FunctionSig{
+		Form:        0,
+		ParamTypes:  []wasm.ValueType{},
+		ReturnTypes: []wasm.ValueType{},
+	}
+
+	// List of all function types available in this module.
+	// There is only one: (func [] -> [])
+	m.Types = &wasm.SectionTypes{
+		Entries: []wasm.FunctionSig{fsig, fsig},
+	}
+
+	m.Function = &wasm.SectionFunctions{
+		Types: []uint32{0, 0},
+	}
+
+	// The body of the start function, that should only
+	// call the host function
+	fb := wasm.FunctionBody{
+		Module: m,
+		Locals: []wasm.LocalEntry{},
+		// code should disassemble to:
+		// call 1 (which is host)
+		// end
+		Code: []byte{0x02, 0x00, 0x10, 0x01, 0x0b},
+	}
+
+	// There was no call to `ReadModule` so this part emulates
+	// how the module object would look like if the function
+	// had been called.
+	m.FunctionIndexSpace = []wasm.Function{
+		{
+			Sig:  &fsig,
+			Body: &fb,
+		},
+		{
+			Sig:  &fsig,
+			Host: reflect.ValueOf(testHostFunction),
+		},
+	}
+
+	m.Code = &wasm.SectionCode{
+		Bodies: []wasm.FunctionBody{fb},
+	}
+
+	// Once called, NewVM will execute the module's main
+	// function.
+	vm, err := NewVM(m)
+	if err != nil {
+		t.Fatalf("Error creating VM: %v", vm)
+	}
+
+	if len(vm.funcs) < 1 {
+		t.Fatalf("Need at least a start function!")
+	}
+
+	// Only one entry, which should be a function
+	if secretVariable != secretValue {
+		t.Fatalf("x is %d instead of %d", secretVariable, secretValue)
+	}
+}
+
+var moduleCallHost = []byte{
+	0x00, 0x61, 0x73, 0x6D, 0x01, 0x00, 0x00, 0x00, 0x01, 0x1A, 0x06, 0x60, 0x01, 0x7F, 0x00, 0x60,
+	0x01, 0x7F, 0x01, 0x7F, 0x60, 0x00, 0x01, 0x7F, 0x60, 0x00, 0x00, 0x60, 0x00, 0x01, 0x7C, 0x60,
+	0x01, 0x7F, 0x01, 0x7F, 0x02, 0x0F, 0x01, 0x03, 0x65, 0x6E, 0x76, 0x07, 0x5F, 0x6E, 0x61, 0x74,
+	0x69, 0x76, 0x65, 0x00, 0x05, 0x03, 0x02, 0x01, 0x02, 0x04, 0x04, 0x01, 0x70, 0x00, 0x02, 0x06,
+	0x10, 0x03, 0x7F, 0x01, 0x41, 0x00, 0x0B, 0x7F, 0x01, 0x41, 0x00, 0x0B, 0x7F, 0x00, 0x41, 0x01,
+	0x0B, 0x07, 0x09, 0x01, 0x05, 0x5F, 0x6D, 0x61, 0x69, 0x6E, 0x00, 0x01, 0x09, 0x01, 0x00, 0x0A,
+	0x08, 0x01, 0x06, 0x00, 0x41, 0x00, 0x10, 0x00, 0x0B,
+}
+
+func add3(x int32) int32 {
+	return x + 3
+}
+
+func importer(name string) (*wasm.Module, error) {
+	m := wasm.NewModule()
+	m.Types = &wasm.SectionTypes{
+		// List of all function types available in this module.
+		// There is only one: (func [int32] -> [int32])
+		Entries: []wasm.FunctionSig{
+			{
+				Form:        0,
+				ParamTypes:  []wasm.ValueType{wasm.ValueTypeI32},
+				ReturnTypes: []wasm.ValueType{wasm.ValueTypeI32},
+			},
+		},
+	}
+	m.FunctionIndexSpace = []wasm.Function{
+		{
+			Sig:  &m.Types.Entries[0],
+			Host: reflect.ValueOf(add3),
+			Body: &wasm.FunctionBody{},
+		},
+	}
+	m.Export = &wasm.SectionExports{
+		Entries: map[string]wasm.ExportEntry{
+			"_native": {
+				FieldStr: "_naive",
+				Kind:     wasm.ExternalFunction,
+				Index:    0,
+			},
+		},
+	}
+
+	return m, nil
+}
+
+func TestHostSymbolCall(t *testing.T) {
+	m, err := wasm.ReadModule(bytes.NewReader(moduleCallHost), importer)
+	if err != nil {
+		t.Fatalf("Could not read module: %v", err)
+	}
+	vm, err := NewVM(m)
+	if err != nil {
+		t.Fatalf("Could not instantiate vm: %v", err)
+	}
+	rtrns, err := vm.ExecCode(1)
+	if err != nil {
+		t.Fatalf("Error executing the default function: %v", err)
+	}
+	if int(rtrns.(uint32)) != 3 {
+		t.Fatalf("Did not get the right value. Got %d, wanted %d", rtrns, 3)
+	}
+}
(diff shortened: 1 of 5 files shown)


gballet:
Good points, going a bit deeper in the spec I [found](https://webassembly.github.io/spec/core/exec/modules.html) that indeed goFunctions should just be a type of `compiledFunctions`. I understand from the spec:

>  Let `hostfunc` be the host function to allocate and `functype` its function type.
> Let `a` be the first free function address in `S`.
> Let `funcins`t be the function instance `{type functype,hostcode hostfunc}`
> Append `funcinst` to the `funcs` of S
> Return `a`

that the host(native) functions should always be at the end of the function index space.

As a result, I will also rename `Native` into `Host`.

dennwc:
Reviewed 5 of 5 files at r3.
Review status: all files reviewed at latest revision, all discussions resolved.

---

*[exec/call.go, line 55 at r3](https://reviewable.io/reviews/go-interpreter/wagon/58#-LCdOigy-t8kQ4iXEndZ:-LCdOigy-t8kQ4iXEnd_:b-9xpapw) ([raw file](https://github.com/go-interpreter/wagon/blob/7afd87873c70c82c9202d49b619a8a33b5793870/exec/call.go#L55)):*
> ```Go
> 		vm.funcs[index].call(vm, int64(index))
> 	} else {
> 		vm.doCall(vm.funcs[index].(compiledFunction), int64(index))
> ```

could it call a `call` function on a compiled function instead?

---

*[exec/call.go, line 89 at r3](https://reviewable.io/reviews/go-interpreter/wagon/58#-LCdP6Sh2jTfsnAOiDNa:-LCdP6Sh2jTfsnAOiDNb:b-2guby3) ([raw file](https://github.com/go-interpreter/wagon/blob/7afd87873c70c82c9202d49b619a8a33b5793870/exec/call.go#L89)):*
> ```Go
> 	}
> 
> 	vm.doCall(vm.funcs[elemIndex].(compiledFunction), int64(elemIndex))
> ```

same here

---

*[exec/vm.go, line 262 at r3](https://reviewable.io/reviews/go-interpreter/wagon/58#-LCdPbUY8WHHNZjP4mV7:-LCdPbUY8WHHNZjP4mV8:b2ziten) ([raw file](https://github.com/go-interpreter/wagon/blob/7afd87873c70c82c9202d49b619a8a33b5793870/exec/vm.go#L262)):*
> ```Go
> 		return nil, ErrInvalidArgumentCount
> 	}
> 	compiled := vm.funcs[fnIndex].(compiledFunction)
> ```

So it cannot run native Go functions here?

---

*[wasm/module.go, line 81 at r3](https://reviewable.io/reviews/go-interpreter/wagon/58#-LCdPuBv0TCkGA5Mxaks:-LCdPuBv0TCkGA5Mxakt:bb7phvt) ([raw file](https://github.com/go-interpreter/wagon/blob/7afd87873c70c82c9202d49b619a8a33b5793870/wasm/module.go#L81)):*
> ```Go
> }
> 
> // ResoclveFunc is a function that takes a module name and
> ```

typo

---


*Comments from [Reviewable](https://reviewable.io/reviews/go-interpreter/wagon/58)*

exec/vm.go
```diff
@@ -80,12 +80,26 @@ func NewVM(module *wasm.Module) (*VM, error) {
...
 	vm.globals = make([]uint64, len(module.GlobalIndexSpace))
 	vm.newFuncTable()
 	vm.module = module
 
+	nNatives := 0
 	for i, fn := range module.FunctionIndexSpace {
+		// Skip native methods as they need not be
+		// disassembled; simply add them at the end
+		// of the `funcs` array as is, as specified
+		// in the spec.
```
sbinet: could you add an explicit pointer to the specs, or the name of the relevant section ?

exec/call_test.go
```diff
@@ -0,0 +1,157 @@
...
+	0x00, 0x61, 0x73, 0x6D, 0x01, 0x00, 0x00, 0x00, 0x01, 0x1A, 0x06, 0x60, 0x01, 0x7F, 0x00, 0x60,
+	0x01, 0x7F, 0x01, 0x7F, 0x60, 0x00, 0x01, 0x7F, 0x60, 0x00, 0x00, 0x60, 0x00, 0x01, 0x7C, 0x60,
+	0x01, 0x7F, 0x01, 0x7F, 0x02, 0x0F, 0x01, 0x03, 0x65, 0x6E, 0x76, 0x07, 0x5F, 0x6E, 0x61, 0x74,
+	0x69, 0x76, 0x65, 0x00, 0x05, 0x03, 0x02, 0x01, 0x02, 0x04, 0x04, 0x01, 0x70, 0x00, 0x02, 0x06,
+	0x10, 0x03, 0x7F, 0x01, 0x41, 0x00, 0x0B, 0x7F, 0x01, 0x41, 0x00, 0x0B, 0x7F, 0x00, 0x41, 0x01,
+	0x0B, 0x07, 0x09, 0x01, 0x05, 0x5F, 0x6D, 0x61, 0x69, 0x6E, 0x00, 0x01, 0x09, 0x01, 0x00, 0x0A,
+	0x08, 0x01, 0x06, 0x00, 0x41, 0x00, 0x10, 0x00, 0x0B,
+}
+
+func testNativeFunctionWithParameterAndReturnValue(x int32) int32 {
```
sbinet: I would just call it `add3` and be done with it :)

exec/call_test.go
```diff
@@ -0,0 +1,157 @@
+package exec
```
sbinet: please add a license header.

exec/call_test.go
```diff
@@ -0,0 +1,157 @@
...
+	testNativeModule.Types = types
+	testNativeModule.FunctionIndexSpace = []wasm.Function{
+		{
+			Sig:        &fsig,
+			Native:     true,
+			NativeBody: reflect.ValueOf(testNativeFunctionWithParameterAndReturnValue),
+			Body:       &wasm.FunctionBody{},
+		},
+	}
+	testNativeModule.Export = exports
```
sbinet: couldn't we create and fill `.Export` in one go?
```go
testNativeModule.Export = &wasm.SectionExports{
    Entries: map[string]wasm.ExportEntry{
        "_native": wasm.ExportEntry{ ... },
    },
}
```

ditto for `testNativeModule.Types`.

exec/call_test.go
```diff
@@ -0,0 +1,157 @@
...
+	vm, err := NewVM(m)
+	if err != nil {
+		t.Fatalf("Could not instantiate vm: %v", err)
+	}
+	rtrns, err := vm.ExecCode(1)
+	if err != nil {
+		t.Fatalf("Error executing the default function: %v", err)
+	}
+	if int(rtrns.(uint32)) != 3 {
+		t.Fatalf("Did not get the right value. Expected %d, got %d", 3, rtrns)
```
sbinet: the convention (in Go, that I tried/try to follow in my projects) is to display first what we got, and then what we wanted.

wasm/module.go
```diff
@@ -20,8 +21,10 @@ const (
 
 // Function represents an entry in the function index space of a module.
 type Function struct {
-	Sig  *FunctionSig
-	Body *FunctionBody
+	Sig        *FunctionSig
+	Native     bool
```
sbinet: couldn't we go with:
```go
type Function struct {
    Sig   *FunctionSig
    Body  *FunctionBody
    Host reflect.Value
}

// IsHost returns whether this function is a host function as defined in:
//  https://webassembly.github.io/spec/core/exec/modules.html#host-functions
func (fct *Function) IsHost() bool {
    return fct.Host != reflect.Value{}
}
```

exec/call_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+import (
+	"bytes"
+	"reflect"
+	"testing"
+
+	"github.com/go-interpreter/wagon/wasm"
+)
+
+func TestNativeCall(t *testing.T) {
```
sbinet: `s/TestNativeCall/TestHostCall/`

exec/call_test.go
```diff
@@ -0,0 +1,157 @@
...
+	const secretValue = 0xdeadbeef
+
+	var secretVariable int = 0
+
+	// a native function that can be called by WASM code.
```
sbinet: `s/native/host/`

exec/call_test.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	// List of all function types available in this module.
+	// There is only one: (func [int32] -> [int32])
+	types := &wasm.SectionTypes{
+		Entries: []wasm.FunctionSig{fsig},
+	}
+
+	exports := &wasm.SectionExports{Entries: make(map[string]wasm.ExportEntry)}
+	exports.Entries["_native"] = wasm.ExportEntry{
```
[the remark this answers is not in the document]
gballet: The C compiler that I used to generate the binary has added an underscore to the symbol. It's not a very strong reason and I *could* remove it, however I expect the underscore to appear in many C-generated WASM files and as such I would like to keep it.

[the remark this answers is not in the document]
sbinet: ok. just idiosyncracies, then.

it's a bit surprising that an exported function starts with `_` (which usually, in many languages, used to denote anything but export/public-ness) :)

but OK.

sbinet:
Reviewed 4 of 4 files at r6.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[exec/call_test.go, line 21 at r4](https://reviewable.io/reviews/go-interpreter/wagon/58#-LBzLcfDdDrn37kdJ40l-r4-21:-LCmAMY-6tizkbVnlDVL:b-266h4m) ([raw file](https://github.com/go-interpreter/wagon/blob/e00af12d0cc8dc6919e52cfd3c7a08f417777e66/exec/call_test.go#L21)):*

`s/testNativeModule/m/`
</blockquote></details>

we are in the `TestHostCall` function.
I still feel it's a bit redundant here to call the module `testHostModule` instead of just `m` or `module`.

what do you think?

---

*[exec/call_test.go, line 92 at r6](https://reviewable.io/reviews/go-interpreter/wagon/58#-LCmBH--2k3O4mu_S551:-LCmBH--2k3O4mu_S552:b-g575bt) ([raw file](https://github.com/go-interpreter/wagon/blob/b4319579ebdff1bb0cbe40cc3108417285a622bf/exec/call_test.go#L92)):*
> ```Go
> }
> 
> var ModuleCallHost = []byte{
> ```

actually, perhaps unexport it?
(I know we are in a `_test.go` file, but...)

---

*[exec/call_test.go, line 107 at r6](https://reviewable.io/reviews/go-interpreter/wagon/58#-LCmAxSSCu5NAP2VKlmn:-LCmAxSSCu5NAP2VKlmo:b-yop8ec) ([raw file](https://github.com/go-interpreter/wagon/blob/b4319579ebdff1bb0cbe40cc3108417285a622bf/exec/call_test.go#L107)):*
> ```Go
> 
> func importer(name string) (*wasm.Module, error) {
> 	testHostModule := wasm.NewModule()
> ```

ditto.

I still feel it's a bit redundant here to call the module `testHostModule` instead of just `m` or `module`.

what do you think?

---

*[exec/vm.go, line 262 at r3](https://reviewable.io/reviews/go-interpreter/wagon/58#-LCdPbUY8WHHNZjP4mV7:-LCmBhq_9dd8qjvnAFIQ:b89hj34) ([raw file](https://github.com/go-interpreter/wagon/blob/7afd87873c70c82c9202d49b619a8a33b5793870/exec/vm.go#L262)):*

So it cannot run native Go functions here?
</blockquote></details>

perhaps do a comma-ok type assertion and panic with some informative message?

---

*[exec/vm.go, line 334 at r6](https://reviewable.io/reviews/go-interpreter/wagon/58#-LCmByXo9KKT0z4s7ufW:-LCmByXo9KKT0z4s7ufX:b-szu6zn) ([raw file](https://github.com/go-interpreter/wagon/blob/b4319579ebdff1bb0cbe40cc3108417285a622bf/exec/vm.go#L334)):*
> ```Go
> 			index := vm.fetchInt64()
> 			label := vm.popInt32()
> 			table := vm.funcs[vm.ctx.curFunc].(compiledFunction).branchTables[index]
> ```

here as well? (comma-ok + panic)

---


*Comments from [Reviewable](https://reviewable.io/reviews/go-interpreter/wagon/58)*

sbinet:
ah... @gballet, there's a conflict now with the just-merged PR #57.
could you fix it?

thanks.

gballet:
resolved + squashed

sbinet:
merged as d4bc452.

thanks!

PR #59, opened 2018-06-04
Pass the VM pointer to host go functions

gballet:
The previous PR #58 was enabling go functions, but without the VM context being passed, they can only do so much. This PR enforces that host Go functions are passed a `*VM` parameter as their first argument.

Another idea would be to check for pointers in the list of arguments, and pass a `*VM` each time that an argument is being encountered. It has some advantages, like streamlining the code and letting the Go type system handle the `panic`s for us. This being said, this leaves too much room for abuse since users could write confusing functions that take that pointer in the middle of their arguments list, thus disturbing the order to the WASM call.

---
This change is [Reviewable](https://reviewable.io/reviews/go-interpreter/wagon/59)

### DIFF
--- a/exec/call_test.go
+++ b/exec/call_test.go
@@ -15,10 +15,10 @@ import (
 func TestHostCall(t *testing.T) {
 	const secretValue = 0xdeadbeef
 
-	var secretVariable int = 0
+	var secretVariable int
 
 	// a host function that can be called by WASM code.
-	testHostFunction := func() {
+	testHostFunction := func(proc *Process) {
 		secretVariable = secretValue
 	}
 
@@ -99,11 +99,11 @@ var moduleCallHost = []byte{
 	0x08, 0x01, 0x06, 0x00, 0x41, 0x00, 0x10, 0x00, 0x0B,
 }
 
-func add3(x int32) int32 {
+func add3(proc *Process, x int32) int32 {
 	return x + 3
 }
 
-func importer(name string) (*wasm.Module, error) {
+func importer(name string, f func(*Process, int32) int32) (*wasm.Module, error) {
 	m := wasm.NewModule()
 	m.Types = &wasm.SectionTypes{
 		// List of all function types available in this module.
@@ -119,7 +119,44 @@ func importer(name string) (*wasm.Module, error) {
 	m.FunctionIndexSpace = []wasm.Function{
 		{
 			Sig:  &m.Types.Entries[0],
-			Host: reflect.ValueOf(add3),
+			Host: reflect.ValueOf(f),
+			Body: &wasm.FunctionBody{},
+		},
+	}
+	m.Export = &wasm.SectionExports{
+		Entries: map[string]wasm.ExportEntry{
+			"_native": {
+				FieldStr: "_naive",
+				Kind:     wasm.ExternalFunction,
+				Index:    0,
+			},
+		},
+	}
+
+	return m, nil
+}
+
+func invalidAdd3(x int32) int32 {
+	return x + 3
+}
+
+func invalidImporter(name string) (*wasm.Module, error) {
+	m := wasm.NewModule()
+	m.Types = &wasm.SectionTypes{
+		// List of all function types available in this module.
+		// There is only one: (func [int32] -> [int32])
+		Entries: []wasm.FunctionSig{
+			{
+				Form:        0,
+				ParamTypes:  []wasm.ValueType{wasm.ValueTypeI32},
+				ReturnTypes: []wasm.ValueType{wasm.ValueTypeI32},
+			},
+		},
+	}
+	m.FunctionIndexSpace = []wasm.Function{
+		{
+			Sig:  &m.Types.Entries[0],
+			Host: reflect.ValueOf(invalidAdd3),
 			Body: &wasm.FunctionBody{},
 		},
 	}
@@ -137,7 +174,7 @@ func importer(name string) (*wasm.Module, error) {
 }
 
 func TestHostSymbolCall(t *testing.T) {
-	m, err := wasm.ReadModule(bytes.NewReader(moduleCallHost), importer)
+	m, err := wasm.ReadModule(bytes.NewReader(moduleCallHost), func(n string) (*wasm.Module, error) { return importer(n, add3) })
 	if err != nil {
 		t.Fatalf("Could not read module: %v", err)
 	}
@@ -151,5 +188,52 @@ func TestHostSymbolCall(t *testing.T) {
 	}
 	if int(rtrns.(uint32)) != 3 {
 		t.Fatalf("Did not get the right value. Got %d, wanted %d", rtrns, 3)
+	}
+}
+
+func TestGoFunctionCallChecksForFirstArgument(t *testing.T) {
+	m, err := wasm.ReadModule(bytes.NewReader(moduleCallHost), invalidImporter)
+	if err != nil {
+		t.Fatalf("Could not read module: %v", err)
+	}
+	defer func() {
+		if r := recover(); r == nil {
+			t.Errorf("This code should have panicked.")
+		} else {
+			if r != "exec: the first argument of a host function was int32, expected ptr" {
+				t.Errorf("This should have panicked because of the wrong type being used as a first argument, and it panicked because of %v", r)
+			}
+		}
+	}()
+	vm, err := NewVM(m)
+	if err != nil {
+		t.Fatalf("Could not instantiate vm: %v", err)
+	}
+	_, err = vm.ExecCode(1)
+	if err != nil {
+		t.Fatalf("Error executing the default function: %v", err)
+	}
+}
+
+func terminate(proc *Process, x int32) int32 {
+	proc.Terminate()
+	return 3
+}
+
+func TestHostTerminate(t *testing.T) {
+	m, err := wasm.ReadModule(bytes.NewReader(moduleCallHost), func(n string) (*wasm.Module, error) { return importer(n, terminate) })
+	if err != nil {
+		t.Fatalf("Could not read module: %v", err)
+	}
+	vm, err := NewVM(m)
+	if err != nil {
+		t.Fatalf("Could not instantiate vm: %v", err)
+	}
+	_, err = vm.ExecCode(1)
+	if err != nil {
+		t.Fatalf("Error executing the default function: %v", err)
+	}
+	if vm.abort == false || vm.ctx.pc > 0xa {
+		t.Fatalf("Terminate did not abort execution: abort=%v, pc=%#x", vm.abort, vm.ctx.pc)
 	}
 }
--- a/exec/func.go
+++ b/exec/func.go
@@ -31,10 +31,20 @@ type goFunction struct {
 }
 
 func (fn goFunction) call(vm *VM, index int64) {
+	// numIn = # of call inputs + vm, as the function expects
+	// an additional *VM argument
 	numIn := fn.typ.NumIn()
 	args := make([]reflect.Value, numIn)
+	proc := NewProcess(vm)
 
-	for i := numIn - 1; i >= 0; i-- {
+	// Pass proc as an argument. Check that the function indeed
+	// expects a *Process argument.
+	if reflect.ValueOf(proc).Kind() != fn.typ.In(0).Kind() {
+		panic(fmt.Sprintf("exec: the first argument of a host function was %s, expected %s", fn.typ.In(0).Kind(), reflect.ValueOf(vm).Kind()))
+	}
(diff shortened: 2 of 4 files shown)


sbinet:
I understand the motivation for introducing this additional parameter (_ie:_ retrieve arguments from the VM, right?) but this seems needlessly constraining.

couldn't we, instead, provide a function:
```go
func WrapFunc(f interface{}) exec.function {
    // create a closure with reflect.FuncOf(...)
    // that closure would have *exec.VM as first arg
}
```

what do you think?

gballet:
@sbinet the motivation is in fact to let host functions that are passed a pointer, to access `memory`.

As far as I can tell, it is not possible to use closures to work around this problem. Consider the following code snippet:

```go
...
m, err := wasm.ReadModule(bufio.NewReader(fd), ModuleResolver)
if err != nil {
	return errors.New(fmt.Sprintf("Could not read module: %v", err))
}

vm, err := exec.NewVM(m)
if err != nil {
	return errors.New(fmt.Sprintf("Could not instantiate vm: %v", err))
}
...
```

`ModuleResolver` does the function resolving, and therefore the "linking" to host functions. It is the one that should create the closure, but won't be able to capture `vm` because at the time it is called, `NewVM` hasn't been called yet.

What I can propose is to "inject" `vm` each time I see a `ptr` in the list of parameters, this way the client code doesn't have to worry about getting a `*VM` object when they don't need it. What do you think?

dennwc:
Totally agree with @gballet - closures won't work here, plus binding the module host function to a single VM sounds weird.

On the other hand, I don't think exposing a VM is a good idea. This way host function will be able to mess with the stack. I propose to define a separate interface that allows to access/allocate the memory and get it as `[]byte` in the host function. It may additionally expose a method to call WASM functions (can be done later).

sbinet:
ok. what @dennwc says SGTM.

but before getting to that, I suppose @gballet could perhaps give us an example of what this PR aims to accomplish or enable. (the tests didn't really exercize the new `exec.VM` argument...)
this way, we'll get a more concrete example+constraints for the interface.

what do you think?

gballet:
@sbinet an example of why host functions would want to access the memory area? At the highest level, it's to ensure that host and wasm functions can act in the same capacity (i.e. access the memory, call other functions, ...) so that the caller doesn't need to know if they're calling a wasm or host function.

If you're asking for a more specific (and therefore contrived) example, I have a host function that takes as parameter a pointer to an input buffer, and a pointer to an output buffer, as well as the maximum length of both buffers. Writing buffers of undefined length is one use case, another one is to write a `BigInt` and a string.

So I would be happy with an interface that goes like:

```golang
type VMInterface interface {
  WriteBuffer(offset int32, maxLength uint32, data []byte) error
  ReadBuffer(offset int32, numBytes uint32, output []byte) error
  // Helpers that are just wrappers around the previous functions
  WriteInt32(offset int32, value int32) error
  ReadInt32(offset int32) (int32, error)
  WriteZeroTerminatedString(offset int32, maxLength uint32, str string) error
  ReadZeroTerminatedString(offset int32) (string, error)
  // For later, it's really just a wrapper around vm.execCode(...)
  callFunction(index uint32, ...) error
}
```

Let me know what you think, and I'd be happy to update the PR with a first draft of that interface.

And if you want me to update the unit tests so that they present the use case beforehand, let me know as well.

dennwc:
@gballet Maybe it's a dumb question to ask, but why each Read and Write includes a Max argument?

Also, I think that there might be a better solution for writing the memory. Imagine a host function that has a very high deviation in terms of input size vs output size, so the caller will not be able to predict an exact buffer size, and allocation the maximal buffer will have a severe performance impact. I think that allocation interface might be better for this purpose. This will allow host functions to allocate the buffer of any size in the VM memory and provide the pointer to the caller.

Something like this:
```golang
type VMInterface interface {
   Make(n int) []byte
   // ...
}
```

It implies a bit of a magic and should be more verbose, like:

```golang
Make(n int) (buf []byte, vmptr uint32)
```

But I think that VM can easily track the buffer that was returned by `Make` and convert it to VM pointer when needed. For example:

```golang
func sayHello(vm VMInterface) []byte {
   s := "hello!"
   b := vm.Make(len(s))
   copy(b, s)
   return b
}
```
It might also be smart enough to get that the buffer returned by the function was not allocated by Make and can do alloc+copy automatically. In this case the host code is even shorter (assuming vm argument is optional):

```golang
func sayHello() []byte {
   return []byte("hello!")
}
```

gballet:
@dennwc all the limits come from my time as an Asm/C/C++ dev, and are probably overkill :smile: 

This being said, I think that having a something like `Make` is dangerous: it gives the user the impression that the allocated memory will be garbage-collected, which it won't be.

My intent is to provide a low-level interface, similar to what you find at the Asm/C level.  I think of host functions like system calls. Garbage-collection is a much higher-level feature that is implemented at runtime-level and differs from one language to the next. This should be left to the runtime, in my opinion.
If the programmer of a host function *really* wants to allocate memory in the VM's memory space, they should use the 'call a WASM function' call that you suggested in a previous post and call the runtime's memory allocation function... It will feel wrong and cumbersome, and therefore dangerous. In general, we should play it safe and not assume too much how the module will manage its memory.

gballet:
One function that I would like to add is `Terminate(msg string)` which stops execution of the VM with an optional reason.

Xe:
I ended up getting around this by having my [Process type wrap an `exec.VM`](https://git.xeserv.us/xena/land/src/00c3cbf4d7b6eb6dc1c18f0886f450b4c8550443/cmd/land/process.go#L21).

exec/vm.go
```diff
@@ -397,3 +399,49 @@ outer:
...
+}
+
+// NewVMInterface creates a VM interface object for host functions
+func NewVMInterface(vm *VM) *VMInterface {
+	return &VMInterface{vm: vm}
+}
+
+// WriteBytes takes an array of bytes and copies it at offset into the
+// module's memory.
+func (vi *VMInterface) WriteBytes(offset int, data []byte) error {
```
sbinet: this method really looks like an `io.WriterAt`:
```go
type WriterAt interface {
	WriteAt(p []byte, off int64) (n int, err error)
}
```

it feels really tempting to implement it... :)

exec/vm.go
```diff
@@ -397,3 +399,49 @@ outer:
...
+	}
+
+	copy(mem[offset:], data)
+	return nil
+}
+
+// ReadBytes returns a slice pointing to a series of byte in memory. The
+// data is copied which is inefficient but if client code direct access
+// to memory this whole interface is moot.
+func (vi *VMInterface) ReadBytes(offset int, length int) ([]byte, error) {
```
sbinet: and this one looks like an `io.ReaderAt`...

gballet:
Actually, according to [what the spec has to say about host functions](https://webassembly.github.io/spec/core/exec/instructions.html#host-functions) they say:

> A host function may also modify the store. However, all store modifications must result in an extension of the original store, i.e., they must only modify mutable contents and must not have instances removed. Furthermore, the resulting store must be valid, i.e., all data and code in it is well-typed.

I feel a better name would be `Store`, although in that case it would not make sense to add `Terminate` to the list of functions. I guess I'll stick to `Process` for now.

exec/vm.go
```diff
@@ -397,3 +399,63 @@ outer:
...
+		length = len(mem) - int(off)
+	} else {
+		length = len(p)
+	}
+
+	copy(p, mem[off:off+int64(length)])
+
+	var err error
+	if length < len(p) {
+		err = fmt.Errorf("Tried to read past the end of memory: wanted to read %d bytes, wrote %d", len(p), length)
```
sbinet: I'd be tempted to reuse `io.ErrShortBuffer` here.

exec/vm.go
```diff
@@ -397,3 +399,63 @@ outer:
...
+		length = len(mem) - int(off)
+	} else {
+		length = len(p)
+	}
+
+	copy(mem[off:], p[:length])
+
+	var err error
+	if length < len(p) {
+		err = fmt.Errorf("Tried to write past the end of memory: wanted to write %d bytes, wrote %d", len(p), length)
```
sbinet: and `io.ErrShortWrite` here.

what do you think?

exec/vm.go
```diff
@@ -397,3 +399,63 @@ outer:
...
 }
+
+// Process is a proxy passed to host functions in order to access
+// things such as memory and control.
+type Process struct {
+	vm *VM
+}
+
+// NewVMInterface creates a VM interface object for host functions
+func NewVMInterface(vm *VM) *Process {
```
sbinet: `s/NewVMInterface/NewProcess/`

exec/func.go
```diff
@@ -31,10 +31,20 @@ type goFunction struct {
...
 func (fn goFunction) call(vm *VM, index int64) {
+	// numIn = # of call inputs + vm, as the function expects
+	// an additional *VM argument
 	numIn := fn.typ.NumIn()
 	args := make([]reflect.Value, numIn)
+	vi := NewVMInterface(vm)
 
-	for i := numIn - 1; i >= 0; i-- {
+	// Pass vi as an argument. Check that the function indeed
+	// expects a *VMInterface argument.
```
sbinet: `s/VMInterface/Process/`

exec/call_test.go
```diff
@@ -99,11 +99,11 @@ var moduleCallHost = []byte{
 	0x08, 0x01, 0x06, 0x00, 0x41, 0x00, 0x10, 0x00, 0x0B,
 }
 
-func add3(x int32) int32 {
+func add3(vi *Process, x int32) int32 {
```
sbinet: `s/vi/proc/` (or `p`)

exec/call_test.go
```diff
@@ -15,10 +15,10 @@ import (
 func TestHostCall(t *testing.T) {
 	const secretValue = 0xdeadbeef
 
-	var secretVariable int = 0
+	var secretVariable int
 
 	// a host function that can be called by WASM code.
-	testHostFunction := func() {
+	testHostFunction := func(vi *Process) {
```
sbinet: `s/vi/proc/` (or `p`)

exec/call_test.go
```diff
@@ -153,3 +190,50 @@ func TestHostSymbolCall(t *testing.T) {
...
+func TestGoFunctionCallChecksForFirstArgument(t *testing.T) {
+	m, err := wasm.ReadModule(bytes.NewReader(moduleCallHost), invalidImporter)
+	if err != nil {
+		t.Fatalf("Could not read module: %v", err)
+	}
+	defer func() {
+		if r := recover(); r == nil {
+			t.Errorf("This code should have panicked.")
+		} else {
+			if r != "exec: the first argument of a host function was int32, expected ptr" {
```
sbinet: this looks brittle.
but I must admit we don't have a safer mechanism in place to handle this kind of thing...

exec/call_test.go
```diff
@@ -153,3 +190,50 @@ func TestHostSymbolCall(t *testing.T) {
...
+				t.Errorf("This should have panicked because of %v", err)
```
sbinet: not sure what `err` will be here.

exec/call_test.go
```diff
@@ -99,11 +99,11 @@ var moduleCallHost = []byte{
 	0x08, 0x01, 0x06, 0x00, 0x41, 0x00, 0x10, 0x00, 0x0B,
 }
 
-func add3(x int32) int32 {
+func add3(vi *Process, x int32) int32 {
 	return x + 3
 }
 
-func importer(name string) (*wasm.Module, error) {
+func importer(name string, f func(*Process, int32) int32) (*wasm.Module, error) {
```
sbinet: we could modify `importer` to take the `f func(*Process, int32) int32)` argument, close around it and return the usual `func(string) (*wasm.Module, error)` importer function...

(just a suggestion. it might make things less obvious, though...)

gballet: I thought of that, and I think this is not the right approach: I can only imagine one real-life use case when a host function needs not access the memory: reading integer value from a system-specific file (like `/proc/$$/fd` or `/dev/urandom`) and that is by no means the most general use case, so it feels like an unnecessary complexity to me. 
Furthermore, the spec specifies that the store should be passed to host functions so it seems normal to do that in any case, just as the spec expects an interpreter to.

sbinet: ok.

exec/call_test.go
gballet: it should be `nil`. That's an interesting mistake, `err` and `r` got mixed up.

gballet: `err` would be nil, and indeed it's not what we should print

sbinet:
merged via f9c0372

PR #80, opened 2018-08-08
Improve sections handling

dennwc:
* Preserve an order of section while decoding the module
* Add a helper to find a custom section by name
* Simplify code paths that read WASM strings by providing a helper
* Support decoding of "name" custom section (debugging information with function names)


wasm/module.go
```diff
@@ -68,44 +69,16 @@ type Module struct {
...
-	if m.Elements != nil {
-		sections = append(sections, m.Elements)
-	}
-	if m.Code != nil {
-		sections = append(sections, m.Code)
-	}
-	if m.Data != nil {
-		sections = append(sections, m.Data)
-	}
+	return m.Sections
```
sbinet: we could perhaps just drop `GetSections()` ?

wasm/module.go
```diff
@@ -68,44 +69,16 @@ type Module struct {
...
+}
+
+func (m *Module) GetCustom(name string) *SectionCustom {
```
sbinet: `s/GetCustom/Custom/` ?

dennwc: I was also thinking to rename `Other` to `Custom` and leave `GetCustom` as-is

dennwc:
Done! Also squashed changes into corresponding commits.

Issue #69, opened 2018-07-18
could not create VM: disasm: stack underflow


valvesss:
Hello, I'm trying to compile and run a simple program from GO to WASM and execute the VM with wagon, but I'm stuck at stack underflow. Any idea?

Code (helloworld.go):

package main
func main(){
  ret := retorno()
  _ = ret
}
func retorno() int64 {
  var num int64 = 50
  return num
}

Compile (With golang beta, go1.11beta1): 

GOARCH=wasm GOOS=js go1.11beta1 build -o test.wasm helloworld.go

Running (wagon functions, latest version) : 

https://gist.github.com/valvesss/c3ac06a597bf25d466271275dd478fb3

Result: 

could not create VM: disasm: stack underflow

evandigby:
I still think it should handle the issue more gracefully, but this isn't working because it needs the "go" module loaded into the runtime (as represented by wasm_exec.js that comes with your Go installation).

If I stub out the go module it gets past this issue and executes the wasm module, meaning I can call "run" which is the entrypoint that Go exposes. The results are less than satisfying because there are a number of imported functions that it uses that are required for it to work.

I'll continue to work on implementing a mirror of the go module:

```go

func main() {
// ...
   g, err := newGoModule()
   if err != nil {
   	log.Fatal("error creating go")
   } 
// ...

	m, err := wasm.ReadModule(f, func(name string) (*wasm.Module, error) {
		switch name {
		case "go":
			return g.m, nil
		default:
			panic("oops" + name)
		}
	})
	if err != nil {
		log.Fatalf("could not read module: %v", err)
	}
}

type gomod struct {
	m *wasm.Module
}

func stub(name string) func(proc *exec.Process, val int32) {
	return func(proc *exec.Process, val int32) { fmt.Println("called", name, "with val", val) }
}

func newGoModule() (*gomod, error) {
	g := &gomod{
		m: wasm.NewModule(),
	}

	g.m.Types = &wasm.SectionTypes{
		Entries: []wasm.FunctionSig{
			{
				Form:       0,
				ParamTypes: []wasm.ValueType{wasm.ValueTypeI32},
			}, // go debug {1} function <func [i32] -> []>
			{
				Form:       0,
				ParamTypes: []wasm.ValueType{wasm.ValueTypeI32},
			}, // go runtime.wasmExit {1} function <func [i32] -> []>
			{
				Form:       0,
				ParamTypes: []wasm.ValueType{wasm.ValueTypeI32},
[...]

PR #81, opened 2018-08-08
disasm: allow to disassemble code without validating it

dennwc:
This CL splits `Disassemble` function into two separate functions: `DisassebleRaw` and `Disasseble`. The first one decodes WASM instructions without checking for validity. And the second function validates the list of instructions from the first function.

This was done to bypass any bugs in the validation code like #69 and be able to manipulate assembly code.


sbinet:
thanks.

I think I'd rather go the other way around:
- make `Disassemble` not decode any wasm instruction
- introduce `DisassembleValidate` (?) that validates the wasm instructions.

perhaps have a `Validate` method on the `Disassembly` type instead of this ugly `DisassembleValidate` function?

Alternatively, we could perhaps introduce a new type, akin to the `encoding/json.Decoder` (_e.g._ `Disassembler`) that has a `Disassemble` method (that would only disassemble, w/o any validation)
that new type could have an exposed field, `Validate bool`, to enable the validation while disassembling.

perhaps the less invasive for now would be something along the lines of:

```go
// Disassemble disassembles a given function body into a set of instructions.
// It won't check operations for validity.
func Disassemble(code []byte) ([]Instr, error) {
    ...
}

type Disassembly struct { ... }

// Disassemble disassembles the given function. It also takes the function's
// parent module as an argument for locating any other functions referenced by
// fn.
func (d *Disassembly) Disassemble(fn wasm.Function, module *wasm.Module) error {
    instrs, err := Disassemble(fn.Body.Code)
    if err != nil { ... }
    // ... what used to be the body of Disassemble ...
   return err
}
```

what do you think?

dennwc:
I think the `Disassemble`+`Validate` is the best option for now.

I like an idea of defining a `Disassembler` and the `Validate` flag, but it will be a bit harder to use because the user will have to set the function signature and the module pointer as well. With `Validate` they will be impossible to miss since they will be listed as arguments.

Populating a `Disassembly` struct is also not very intuitive.

sbinet:
sorry... I have been distracted (by apache/go-arrow)

I am not sure I completely get what you're advocating for (sorry!).

point noted about populating `Dissambly` not being too intuitive.

what about:
```go
// Disassemble disassembles a given function body into a set of instructions.
// It won't check operations for validity.
func Disassemble(code []byte) ([]Instr, error) {
    ...
}

// NewDisassembly disassembles the given function. It also takes the function's
// parent module as an argument for locating any other functions referenced by
// fn.
func NewDisassembly(fn wasm.Function, module *wasm.Module) (*Disassembly, error) {
    instrs, err := Disassemble(fn.Body.Code)
    if err != nil { ... }
    // ... what used to be the body of Disassemble ...
   return err
}
```

dennwc:
@sbinet Changed according to your last suggestion. PTAL

disasm/disasm_test.go
```diff
@@ -0,0 +1,41 @@
+package disasm_test
```
sbinet: Please add a copyright header.

sbinet:
Meh... I can add the header later :)

PR #84, opened 2018-08-22
wasm: allocate module.TableIndexSpace when importing table from resolver

duanbing:
When I run this case  [math](https://github.com/duanbing/wasm-test/blob/master/go/math.go), I got an error as below:
```
panic: runtime error: index out of range

goroutine 1 [running]:
github.com/go-interpreter/wagon/wasm.(*Module).resolveImports(0xc4200a6120, 0x115b720, 0x1, 0xc4200842c0)
	/Users/XXXX/Project/Go/src/github.com/go-interpreter/wagon/wasm/imports.go:187 +0xcd2
github.com/go-interpreter/wagon/wasm.ReadModule(0x116fd00, 0xc42007c6c0, 0x115b720, 0x13e, 0x33e, 0x0)
	/Users/XXXX/Project/Go/src/github.com/go-interpreter/wagon/wasm/module.go:143 +0x2a8
main.main()
	/Users/XXXX/Project/Go/src/github.com/duanbing/wasm-test/go/math.go:33 +0xf5
exit status 2
```

After tracing the source code, I found `module.TableIndexSpace` in file `wasm/imports.go line 187`, didn't get allocated. 
   
So I went back to function `DecodeModule`  in `ReadModule` to figure out whether module.TableIndexSpace is allocated, or `m.Table` is allocated.    

In `ReadModule`, we firstly `DecodeModule math.wasm`,  coz I don't declare table, so `m.TableIndexSpace` is not be allocated, but m.LinearMemoryIndexSpace is.   Then go to `resolveImports`, travel `module.Import.Entries`, which is Import Section of `math.wasm`,  and use Export Section from env.wasm to fill in those sections. Unfortunately,  when  initializing m.TableIndexSpace, we got panic.

So, this is all I caught.  I have no idea if it's a bug, or my env.wasm is not written correctly.


sbinet:
Looks good.

2 things:
- could you add a test that would fail w/o this PR?
- could you reword your commit message so its header looks like:
```
wasm: allocate module.TableIndexSpace when importing table from resolver
```

thanks.

duanbing:
I had changed the commit message header.  

For the exception test, I use an `import` instruction. AFAIK,  there is not `import` unit test except for some examples in `vm_example_test`, but I don't think this PR is an example.   And All tests in  testdata/modules.json seems are designed for the single module.  Do you have any suggestions for the test's placement?

PR #88, opened 2018-09-02
Implement WASM text format writer

dennwc:
Implement WASM text format writer with (nearly) the same output as `wasm2wat`.


wast/write_test.go
```diff
@@ -0,0 +1,67 @@
...
+				}
+				buf := new(bytes.Buffer)
+				err = wast.WriteTo(buf, m)
+				if err != nil {
+					t.Fatal(err)
+				}
+				if !bytes.Equal(exp, buf.Bytes()) {
+					ioutil.WriteFile(tname+"_got", buf.Bytes(), 0644)
+					t.Fatalf("output is different")
+				} else {
```
sbinet: please remove the `else` branch and block.
the `tname+"_got"` is only written out when there's a failure.

wast/write.go
```diff
@@ -0,0 +1,373 @@
...
+)
+
+const tab = `  `
+
+// WriteTo writes a WASM modules in a text representation.
+func WriteTo(w io.Writer, m *wasm.Module) error {
+	var fnames wasm.NameMap
+	if s := m.Custom(wasm.CustomSectionName); s != nil {
+		var names wasm.NameSection
+		_ = names.UnmarshalWASM(bytes.NewReader(s.Data))
```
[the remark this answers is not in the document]
dennwc: This section is explicitly described as a custom section, and the spec says something along the lines: "tools should be able to read the module even if this section contains random data". This section only adds human-readable names and nothing more, so it's totally safe to ignore.

wast/write.go
```diff
@@ -0,0 +1,373 @@
...
+		sub, _ := names.Decode(wasm.NameFunction)
```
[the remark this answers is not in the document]
dennwc: The same as above. This section is custom and only serves as a hint, thus it should not fail the whole process of printing the module.

wast/write_test.go
dennwc: Actually, the code is totally valid since FS is persistent.

If the test fails, it will write a `_got` file to the same directory. Afterward, the test case can be fixed and the test will clean up an old file.

dennwc:
@sbinet PTAL

wast/write_test.go
```diff
@@ -0,0 +1,67 @@
...
+// license that can be found in the LICENSE file.
+
+package wast_test
+
+import (
+	"bytes"
+	"io/ioutil"
+	"path/filepath"
+	"testing"
+
```
sbinet: please remove this blank line.

wast/write_test.go
sbinet: ok. I don't really like having state being carried away across compilation/tests, but ok.

wast/write_test.go
dennwc: Done!

sbinet: thanks :)

PR #90, opened 2018-10-16
Warn when modules are missing before attempting disassembly

gballet:
The goal is to catch the error early, otherwise it will cause a very confusing `disasm: stack underflow` which makes debugging very difficult.

### DIFF
--- a/disasm/disasm.go
+++ b/disasm/disasm.go
@@ -98,8 +98,12 @@ func NewDisassembly(fn wasm.Function, module *wasm.Module) (*Disassembly, error)
 	}
 	disas := &Disassembly{}
 
-	if len(module.FunctionIndexSpace) != len(module.Import.Entries)+len(module.Function.Types) {
-		panic(fmt.Sprintf("Attempting to disassemble a module that is missing imports %d != %d + %d", len(module.FunctionIndexSpace), len(module.Import.Entries), len(module.Function.Types)))
+	if module.Import != nil {
+		ninternal, _ := leb128.ReadVarint32(bytes.NewReader(module.Code.Bytes))
+		nexternal := len(module.FunctionIndexSpace) - int(ninternal)
+		if nexternal != len(module.Import.Entries) {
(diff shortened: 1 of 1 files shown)


sbinet:
thanks, but it seems this breaks a bunch of tests.

gballet:
It seems to trigger a lot of issues that I missed out, indeed. I'm going to fix those in several commits over the next few days.

cmd/wasm-dump/main.go
```diff
@@ -82,7 +82,42 @@ func process(w io.Writer, fname string) {
...
+					})
+				}
+			}
+		}
+		return m, nil
+	}
+
+	// We already read the file without trying to resolve imports,
+	// start over in order to do that work now.
+	f.Seek(0, os.SEEK_SET)
```
sbinet: `s/os.SEEK_SET/io.SeekStart/`

cmd/wasm-dump/main.go
```diff
@@ -82,7 +82,42 @@ func process(w io.Writer, fname string) {
 	}
 	defer f.Close()
 
-	m, err := wasm.ReadModule(f, nil)
+	// Decode the module in order to fake its imports as exports
+	mod, err := wasm.DecodeModule(f)
+	if err != nil {
+		log.Fatalf("Could not read module: %v", err)
```
sbinet: `s/Could/could/`

disasm/disasm.go
```diff
@@ -97,6 +98,14 @@ func NewDisassembly(fn wasm.Function, module *wasm.Module) (*Disassembly, error)
 	}
 	disas := &Disassembly{}
 
+	if module.Import != nil {
+		ninternal, _ := leb128.ReadVarint32(bytes.NewReader(module.Code.Bytes))
+		nexternal := len(module.FunctionIndexSpace) - int(ninternal)
+		if nexternal != len(module.Import.Entries) {
+			panic(fmt.Sprintf("Attempting to disassemble a module that is missing imports %d != %d", nexternal, len(module.Import.Entries)))
```
sbinet: `s/fmt.Sprintf/fmt.Errorf/`

PR #92, opened 2018-10-28
don't panic when reading a module without functions; return error

gballet:
When attempting to load a module that doesn't have any function, wagon will panic when trying to resolve functions.This simple patch returns an error at resolve time instead.


gballet:
Thinking about it, I can already see one issue with this fix: if a module declares no functions but simply exports some of the functions it imports, then this fix will return an error. I should instead make sure that if `m.Code` is `nil` then it should be declared empty. What do you think @sbinet ?

PR #93, opened 2018-12-04
fix data init order

gballet:
WASM files declaring overlapping data sections would not end up with the proper state as accepted by binaryen.

This PR simply makes sure that the newly created memory is first initialized with the previous content of the memory before it is overwritten with the content from the next data segment.


gballet:
@sbinet done

sbinet:
Thanks!

PR #96, opened 2018-12-19
wasm: improved error type when check import signature

kr1sten0:
This PR does a little bit improvement of `InvalidImportError ` error type, since only the function signature type index is tip insufficient and a bit confusing

/cc @gballet


kr1sten0:
https://github.com/go-interpreter/license/pull/9 
PR sent. thanks!

PR #104, opened 2019-02-20
disasm: fix if-else bug causing stack underflow error

cdetrio:
Here is a test case to trigger a bug (wasm file included):
```
; ifelse-stack-bug.wast
(module
  (func $main (export "main") (result i32)
    (local $l0 i32)
    (block
      (i32.const 1)
      (i32.const 2)
      (i32.const 3)
      (set_local $l0
        (if (result i32)
          (i32.eq
            (i32.const 1)
            (i32.const 0))
          (then
            (i32.const 65))
          (else
            (i32.const 66))))
      (drop)
      (drop)
      (drop))
    (get_local $l0)
  )
)
```

And here is the trace with `SetDebugMode(true)` in `disasm/log.go`:

```
$ ./cmd/wasm-run/wasm-run -v ./exec/testdata/ifelse-stack-bug.wasm
section.go:114: Reading section ID
section.go:123: Reading payload length
section.go:130: Section payload length: 5
section.go:146: section type
section.go:114: Reading section ID
section.go:123: Reading payload length
section.go:130: Section payload length: 2
section.go:154: section function
section.go:114: Reading section ID
section.go:123: Reading payload length
section.go:130: Section payload length: 8
section.go:170: section export
section.go:114: Reading section ID
section.go:123: Reading payload length
section.go:130: Section payload length: 42
section.go:182: section code
section.go:797: 1 function bodies
section.go:800: Reading function 0
section.go:855: bodySize: 40, localCount: 1
section.go:858: Read 37 bytes for function body
section.go:114: Reading section ID
section.go:123: Reading payload length
section.go:130: Section payload length: 23
section.go:141: section custom
section.go:114: Reading section ID
module.go:165: There are 1 entries in the function index space.
disasm.go:113: stack top is 0
disasm.go:132: op: block, unreachable: false
disasm.go:227: if, depth is 0
disasm.go:113: stack top is 0
disasm.go:132: op: i32.const, unreachable: false
disasm.go:113: stack top is 1
disasm.go:132: op: i32.const, unreachable: false
disasm.go:113: stack top is 2
disasm.go:132: op: i32.const, unreachable: false
disasm.go:113: stack top is 3
disasm.go:132: op: i32.const, unreachable: false
disasm.go:113: stack top is 4
disasm.go:132: op: i32.const, unreachable: false
disasm.go:113: stack top is 5
disasm.go:132: op: i32.eq, unreachable: false
disasm.go:113: stack top is 4
disasm.go:132: op: if, unreachable: false
disasm.go:227: if, depth is 3
disasm.go:113: stack top is 3
disasm.go:132: op: i32.const, unreachable: false
disasm.go:113: stack top is 4
disasm.go:132: op: else, unreachable: false
disasm.go:205: discard 1 elements, preserve top: true
disasm.go:210: setting new stack for if block (7)
disasm.go:113: stack top is 0
disasm.go:132: op: i32.const, unreachable: false
disasm.go:113: stack top is 1
disasm.go:132: op: end, unreachable: false
wasm-run: could not create VM: disasm: stack underflow
```

The underflow error is thrown here, because `curDepth` is 1 and `prevDepth` is 3: https://github.com/go-interpreter/wagon/blob/fde98f722500fce2af8b2b864124bbbce1a1688e/disasm/disasm.go#L196-L198

`prevDepth` is the number of stack elements before entering the if block. The problem is that the stack depth when entering the if-else block should be the same as when entering the if-then block (3 stack elements). The bug is pushing a stack depth of 0 when entering the if-else block. The stack depth when entering the if-else block is the same as when entering the if-then block, so the fix is to push the same stack depth for if-else as done for if-then.


sbinet:
needs https://github.com/go-interpreter/license/pull/13

sbinet:
thanks a lot!

Issue #101, opened 2019-01-29
Would you accept PRs to implement a JIT?


twitchyliquid64:
Heya,

I would like to implement a JIT, starting with a subset of wasm instructions on amd64. Would you be open to accepting PRs?

To start with, the JIT would rewrite sequences of mathy instructions into a unit, which is referenced by PC. An unused opcode can be used to symbolize an execution of a jitted section.

For example:

```
LOAD i32
LOAD i32
ADD
LOAD i32
MUL
JMP blah
```
Could be rewritten to:

```
EXECJIT
NOP
NOP
NOP
NOP
JMP blah
```

`VM` would have a new `map` field, which would store JIT units using the PC as the index. Instructions which are rewritten into JIT code and rewritten to NOPs so the PC jumps remain the same.

Sufficiently long sequences of mathy code (which contain no branches and no inbound branches) will be JITed out, the rest left as-is.

This should eliminate mainloop + Go overhead for really tight, mathematics operations (such as hash computation). Once this works, I start supporting more opcodes, which will in turn mean more execution can be JIT accelerated. Given the VM already has a separate stack, I can avoid using any stack space (and hence avoid messing with Go's internal stack maps + management).

My testbest JIT can do this in linear time, using a greedy register allocation scheme.

Does this sound good to you?

sbinet:
sounds really interesting!
(JITing Go is something I've always wanted to investigate...)

twitchyliquid64:
Calling into JITed code is actually really easy, you just map in an executable page(s) and jump to it:

```go
fp := *(*func(uint64) uint64)(unsafe.Pointer(&f))
return fp(argument)
```

If the runtime happens to do a GC run, it wont recognize the current PC and will try again later. The only thing we have to worry about is stack overrun. I've also confirmed that planned pre-emptive scheduling will do the same thing.

We are guaranteed to have 720 bytes spare as long as we dont have any defers in our calling function (this is the red zone - a spare space buffer that the runtime maintains). If we need more, we can call into `runtime.morestack`, a check+call that appears in every Go function prologue.

The only hard bit left is generating the instructions & register selection. Ive pulled out the Go assembler into a library [here](https://github.com/twitchyliquid64/golang-asm), and have a reasonable linear-time JIT implemented in a private repo (which i will open eventually).

twitchyliquid64:
Pinging this thread. I'm seeking explicit approval @sbinet to implement this, and confirmation that PRs will be reviewed & eventually merged.

[...]

PR #105, opened 2019-03-10
Part 1/2 #101 - implement partial compilation to native code

twitchyliquid64:
This PR lays the boilerplate for supporting AOT & JIT compilation.

- Use one of the unassigned opcodes to signal to the interpreter to execute a native code section. Modify the disassembler to throw an error if this opcode is used, so any collision years down the line is detected & can be trivially fixed.

- Implement interfaces to abstract different stages, specifically:

  1. Scanning opcodes to determine candidates for native compilation + gathering statistics (`sequenceScanner`)
  2. Allocating executable pages, as such an operation is OS dependent (`pageAllocator`)
  3. Actually compiling the chosen sequence into instructions (`instructionBuilder`)

 - When a sequence of opcodes is rewritten, the beginning is set to an internal opcode that signals the interpreter should run a section of native code at a specific `[]asmBlock` index. Any remaining opcodes are set to `ops.Unreachable` - as we dont plan to emit asm where there are inbound branches in the middle (yet), this should crash out in the event of any bugs that jump into the recompiled section.

- Tests for integration between the different interfaces & structures in `VM` / `compiledFunction`.


twitchyliquid64:
On second thought, hold off on reviewing this. I need to think about a better way to structure this.

twitchyliquid64:
@sbinet, can you look at this and give me your thoughts? I'm not done, but getting close. This PR is mostly boilerplate, but usefully implements AOT compilation for `i64.const`, `i64.add`, & `i64.sub`.

 - Changes to the `disasm` package to track the encoded size of each instruction it reads. I'm not using this anymore, but the code + tests are written. Let me know if you want me to leave it in or delete the code.

 - Changes to `exec/exec_test.go` to run each test suite twice - once with AOT compilation disabled & once enabled, all in parallel.

 - Changes to `exec/internal/compile/compile.go` to track instruction metadata and jump targets. `Compile()` now additionally returns a `compile.BytecodeMetadata` object to track information about the compiled bytecode. I use it a little now (mostly to avoid the overhead I would otherwise need to disassemble again), but it will be really invaluable for when we implement a JIT (as opposed to just AOT).

 - Changes to `exec/func.go`:  A slice of emitted native code sections is kept for each function (`compiledFunction.asm`).

 - `exec/internal/compile/allocator.go`: Implements a naive executable page allocator. I know this is very leaky for the moment.

 - `exec/internal/compile/amd64.go`: Implements the backend for the amd64 architecture. This is very simplistic ATM and only supports 3 opcodes, but this is a good start. There are lots of easy optimizations we can make here too - tracking if a register already has stack information or what would be pushed to the stack is a good start. Peephole optimizations and dead-instruction deletion would also be great too.

 - `exec/internal/compile/amd64_test.go`: Test suite for the amd64 backend. Of particular note is the `TestSliceMemoryLayoutAMD64` function, which confirms that the memory layout of the slice (which is critical for the emitted instructions to work).

 - `exec/internal/compile/scanner.go`: Implements a generic scanner which any backend can use. This calculates heuristics about the instructions (used for deciding whether to native-compile them or not, not very useful now but will be for a JIT) as well as working out which subsets have only instructions which are fully supported by the backend.

 - `exec/native_compile.go`: Abstracts access to the native backend, as well as having the main logic for AOT compilation & actually executing native code.

 - `wasm/operators/wagon_internal.go`: I've added a new interpreter opcode to symbolize execution of a sequence of native code.

exec/func.go
```diff
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"math"
 	"reflect"
+	"unsafe"
```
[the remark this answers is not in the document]
twitchyliquid64: Hmmmm, I didnt think of that. Maybe we could wrap `asmBlock` and methods to call into it in its own type, which we can put behind build tags?

disasm/disasm.go
```diff
@@ -388,104 +392,122 @@ func Disassemble(code []byte) ([]Instr, error) {
...
-			res, err := leb128.ReadVarUint32(reader)
+			res, size, err := leb128.ReadVarUint32Size(reader)
 			if err != nil {
 				return nil, err
 			}
 			instr.Immediates = append(instr.Immediates, uint8(res))
+			cursor += size
 		}
+
+		instr.EncodedLen = int(cursor - startingCursor)
```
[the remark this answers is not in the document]
twitchyliquid64: Sure, happy to do that (my preference is `bytes.Reader.Len()`).

Changes to `disasm` like this were made for a previous iteration - I no longer use `EncodedLen` anywhere in the compiler. I am happy to make these changes or revert these files either way.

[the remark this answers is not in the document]
sbinet: ok, then best to remove that field and revert to what `disasm/disasm.go` was :)

exec/func.go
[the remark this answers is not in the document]
sbinet: either build tags or some kind of interface-based code that is passed by the user.

I'd err on the latter mechanism, if possible.

(PS: please rebase on the latest `master` to pick-up the Go-1.12 CI-travis stuff.)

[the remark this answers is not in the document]
twitchyliquid64: I don't see how we could do the latter without exposing the native internals to potential security issues in user code (in any case, the native stuff is in an internal package, so a major refactor would be needed to enable this). I'm more a fan of splitting this into build tags.

disasm/disasm.go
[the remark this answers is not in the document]
twitchyliquid64: Done :)

exec/func.go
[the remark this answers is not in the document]
twitchyliquid64: I refactored a bunch of the native stuff into interfaces, and put all the unsafe stuff behind the tag `appengine`. It builds + tests successfully with that tag applied.

Once this is in can you make the `.travis.yml` change? I had a stab at it but it didnt work.

twitchyliquid64:
Ive addressed the comments, PTAL.

twitchyliquid64:
Friendly ping @sbinet :)

exec/internal/compile/amd64_test.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+}
+
+func TestSliceMemoryLayoutAMD64(t *testing.T) {
+	if runtime.GOARCH != "amd64" {
+		t.SkipNow()
+	}
+
+	slice := make([]uint64, 2, 5)
+	mem := (*[24]byte)(unsafe.Pointer(&slice))
```
sbinet: ah! I thought this was testing something that's a Go1 compatibility guarantee via the `reflect.SliceHeader`, but actually:
```go
// SliceHeader is the runtime representation of a slice.
// It cannot be used safely or portably and its representation may
// change in a later release.
// Moreover, the Data field is not sufficient to guarantee the data
// it references will not be garbage collected, so programs must keep
// a separate, correctly typed pointer to the underlying data.
type SliceHeader struct {
	Data uintptr
	Len  int
	Cap  int
}
```

we should perhaps add some more comments here for the lucky person who will have to decipher why this is failing in Go2 or Go3...

exec/internal/compile/compile.go
```diff
@@ -154,6 +176,7 @@ func Compile(disassembly []disasm.Instr) ([]byte, []*BranchTable) {
 			instr.Immediates = []interface{}{instr.Immediates[1].(uint32)}
 		case ops.If:
 			curBlockDepth++
+			emitMetadata(OpJmpZ, buffer.Len(), 9)
```
sbinet: the size is almost always `9`.
perhaps add a `const` with a nice name that explains why and what it is?
and use that everywhere?
(there's only one occurence of a size of 18.)

exec/internal/compile/native_exec.go
```diff
@@ -0,0 +1,19 @@
+// Copyright 2017 The go-interpreter Authors.  All rights reserved.
```
sbinet: `s/2017/2019/`

exec/internal/compile/amd64_test.go
```diff
@@ -0,0 +1,177 @@
+// Copyright 2017 The go-interpreter Authors.  All rights reserved.
```
sbinet: 2019

exec/internal/compile/amd64.go
```diff
@@ -0,0 +1,271 @@
+// Copyright 2017 The go-interpreter Authors.  All rights reserved.
```
sbinet: 2019

exec/internal/compile/allocator_test.go
```diff
@@ -0,0 +1,45 @@
+// Copyright 2017 The go-interpreter Authors.  All rights reserved.
```
sbinet: 2019

exec/internal/compile/allocator.go
```diff
@@ -0,0 +1,67 @@
+// Copyright 2017 The go-interpreter Authors.  All rights reserved.
```
sbinet: 2019

exec/internal/compile/amd64.go
```diff
@@ -0,0 +1,271 @@
...
+
+package compile
+
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+
+	ops "github.com/go-interpreter/wagon/wasm/operators"
+	asm "github.com/twitchyliquid64/golang-asm"
```
sbinet: please add this new dependency to the `go.mod` file.

exec/internal/compile/allocator.go
```diff
@@ -0,0 +1,67 @@
...
+// license that can be found in the LICENSE file.
+
+// +build !appengine
+
+package compile
+
+import (
+	"unsafe"
+
+	mmap "github.com/edsrzf/mmap-go"
```
sbinet: please add this new dependency to the `go.mod` file.

exec/internal/compile/amd64.go
```diff
@@ -0,0 +1,271 @@
...
+	prog.From.Type = obj.TYPE_REG
+	prog.From.Reg = x86.REG_R9
+	prog.To.Type = obj.TYPE_REG
+	prog.To.Reg = x86.REG_AX
+	switch op {
+	case ops.I64Add:
+		prog.As = x86.AADDQ
+	case ops.I64Sub:
+		prog.As = x86.ASUBQ
+	}
```
sbinet: add a `default` case to catch mistakes?

exec/internal/compile/amd64.go
```diff
@@ -0,0 +1,271 @@
...
+	// cmd.Stdin = bytes.NewReader(out)
+	// cmd.Stdout = os.Stdout
+	// cmd.Run()
+	return out, nil
+}
+
+func (b *AMD64Backend) readIntImmediate(code []byte, meta InstructionMetadata) (uint64, error) {
+	if meta.Size == 5 {
+		var out uint32
+		err := binary.Read(bytes.NewReader(code[meta.Start+1:meta.Start+meta.Size]), binary.LittleEndian, &out)
```
sbinet: I'd use the faster `binary.LittleEndian.Uint32(code[...])` instead.
no need for an `error` in the signature, then.

exec/internal/compile/amd64.go
```diff
@@ -0,0 +1,271 @@
...
+		return uint64(out), err
+	}
+	var out uint64
+	err := binary.Read(bytes.NewReader(code[meta.Start+1:meta.Start+meta.Size]), binary.LittleEndian, &out)
```
sbinet: ditto, with `binary.LittleEndian.Uint64`.

exec/native_compile.go
```diff
@@ -0,0 +1,144 @@
+// Copyright 2017 The go-interpreter Authors.  All rights reserved.
```
sbinet: 2019

exec/internal/compile/scanner.go
```diff
@@ -0,0 +1,108 @@
+// Copyright 2017 The go-interpreter Authors.  All rights reserved.
```
sbinet: 2019

exec/internal/compile/scanner.go
```diff
@@ -0,0 +1,108 @@
...
+	// Bytecode index of the first opcode.
+	Beginning uint
+	// Bytecode index of the last byte in the instruction.
+	End uint
+	// InstructionMeta index of the first instruction.
+	StartInstruction int
+	// InstructionMeta index of the last instruction.
+	EndInstruction int
+	// Metrics about the instructions between first & last index.
+	Metrics *Metrics
```
sbinet: it seems we always expect `Metrics` to be there.
why don't we make it `Metrics` instead of `*Metrics` ?

exec/internal/compile/scanner.go
```diff
@@ -0,0 +1,108 @@
...
+		// can support that in the future.
+		isInsideBranchTarget := meta.InboundTargets[int64(inst.Start)] && inst.Start > 0
+
+		if !s.supportedOpcodes[inst.Op] || isInsideBranchTarget {
+			// See if the candidate can be emitted.
+			if inProgress.Metrics.AllOps > 2 {
+				finishedCandidates = append(finishedCandidates, inProgress)
+			}
+			// Reset the candidate.
+			inProgress = CompilationCandidate{Metrics: &Metrics{}}
```
sbinet: perhaps introduce a `reset()` method instead?

exec/native_compile_test.go
```diff
@@ -0,0 +1,186 @@
+// Copyright 2017 The go-interpreter Authors.  All rights reserved.
```
sbinet: 2019

exec/native_compile_test.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	if err := vm.tryNativeCompile(); err != nil {
+		t.Fatalf("tryNativeCompile() failed: %v", err)
+	}
+
+	// Our scanner emitted two sequences. The first should not have resulted in
+	// compilation, but the second should have. Lets check thats the case.
+	fn := vm.funcs[0].(compiledFunction)
+	if want := 1; len(fn.asm) != want {
```
sbinet: I'd write this as:
```go
if got, want := len(fn.asm), 1; got != want { ...
```

exec/vm.go
```diff
@@ -70,17 +71,30 @@ type VM struct {
...
+type VMOptions struct {
+	// EnableAOT enables ahead-of-time compilation of supported opcodes
+	// into runs of native instructions, if wagon supports native compilation
+	// for the current architecture.
+	EnableAOT bool
+}
+
+// NewVMWithOptions creates a new VM from a given module and options. If the module defines
+// a start function, it will be executed.
+func NewVMWithOptions(module *wasm.Module, options *VMOptions) (*VM, error) {
```
sbinet: I'd rather use the "functional options" pattern than this exported `VMOptions` struct.

_e.g._
```go
type VMOption func(c *config)

func EnableAOT(v bool) VMOption {
    return func(c *config) {
        c.EnableAOT = v
    }
}

func NewVM(module *wasm.Module, opts ...VMOption) (*VM, error) {
...
}
```

what do you think?

wasm/operators/wagon_internal.go
```diff
@@ -0,0 +1,20 @@
+// Copyright 2017 The go-interpreter Authors.  All rights reserved.
```
sbinet: 2019

exec/vm.go
```diff
@@ -47,6 +47,7 @@ type context struct {
 	stack   []uint64
 	locals  []uint64
 	code    []byte
+	asm     *[]asmBlock
```
sbinet: could you explain why the extra indirection?

exec/internal/compile/amd64.go
```diff
@@ -0,0 +1,271 @@
+// Copyright 2017 The go-interpreter Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
```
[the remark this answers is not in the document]
twitchyliquid64: In my next PR I'll enable support for windows - it should work right now, I just havent tested it.

The current architecture of `exec.supportedNativeArchs` is such that all the backend arches need to be compiled in - runtime logic decides which to use. I can re-architect this to just use build tags if you want?

exec/vm.go
twitchyliquid64: Good idea!

exec/vm.go
twitchyliquid64: Nope! fixed.

twitchyliquid64:
Done :) PTAL.

twitchyliquid64:
Apologies for the additional commit - I was intending for that to be a different PR, but I found a few bugs that shouldn't be left out of this PR (off-by-one in the return instruction pointer, double-allocation/overrun for `vm.context.stack`).

twitchyliquid64:
Friendly ping @sbinet :)

twitchyliquid64:
Let me know if there's anything I can do to move this along. If you would like me to split some of this into separate PRs, (like the scanner in one and the execution in another) I can try. Let me know :)

sbinet:
Apologies for the sluggish answers. (I am on the road to the dotGo conference...)

If you can split some PRs off, then yes, by all means: smaller bites are great :)

twitchyliquid64:
No worries at all :) I'm sorry for being pushy, just trying to get this in before my excitement shifts to other ideas.

I'll go ahead and split this into 2-3 different PRs, and close this one.

Thanks, and enjoy dotGo! Say hi to Dave for me :smile:

PR #107, opened 2019-03-11
wasm/leb128: drop ReadXYZSize functions

sbinet:


twitchyliquid64:
Note: These are exported functions, so anyone already using it will have their code break. Could we just add a comment to mark them as deprecated instead?

sbinet:
you're right but:

- `wagon` isn't `v1.0.0` yet so we are "allowed" to break APIs
- going to godoc and listing all importers of `leb128` there's only one package that imports `leb128` and uses the removed function: an old fork of `wagon`: https://github.com/netvm/wagon/blob/master/wasm/section.go#L112

PR #109, opened 2019-03-17
Implement shift+XOR instructions for amd64 native backend.

twitchyliquid64:
This PR relies on #105 - please hold off on review till the former is ready.

This adds support for a few more instructions that are frequently used in math heavy computations (the kind that benefits greatly from AOT compilation).


twitchyliquid64:
Rewritten into #115.

PR #111, opened 2019-03-22
1/4 - Implementent native compilation - Fix stack allocation & depths, add debug mode to find these issues.

twitchyliquid64:
Sooooo I found an issue where the stack size is wrong (two situations - calling into a `compiledFunction`, and when the function returns a value onto the parents stack). This was masked by the use of `append`, which transparently reallocates the slice.

This PR fixes this, and introduces `debugStackDepth` (which compiles away to nothing when disabled).


twitchyliquid64:
Friendly ping on this PR and the others :) Hope dotGo went well :D

.travis.yml
```diff
@@ -35,6 +35,7 @@ sudo: false
 script:
  - go get -d -t -v ./...
  - go install -v $TAGS ./...
+ - sed -i "s/\(const debugStackDepth = \)false/\1true/" ./exec/vm.go
```
sbinet: couldn't we instead use build-tags?
we could put the `debugStackDepth=true` in a, e.g., `exec/vm_debug_stack.go` file with some build-tag (say `debugstack`) and the one with `=false` in `exec/vm_stack.go` (with `!debugstack`)

exec/vm.go
```diff
@@ -18,6 +18,8 @@ import (
 	ops "github.com/go-interpreter/wagon/wasm/operators"
 )
 
+const debugStackDepth = false
```
sbinet: please document what it does for our future selves :)

exec/func.go
```diff
@@ -80,7 +80,7 @@ func (fn goFunction) call(vm *VM, index int64) {
 }
 
 func (compiled compiledFunction) call(vm *VM, index int64) {
-	newStack := make([]uint64, compiled.maxDepth)
+	newStack := make([]uint64, 0, compiled.maxDepth+1)
```
sbinet: please document why we need the extra slot.

twitchyliquid64:
:smiley:

twitchyliquid64:
Thanks :)

Please see https://github.com/go-interpreter/license/pull/14.

PR #112, opened 2019-03-24
2/4 - Implement native compilation - Implement boilerplate.

twitchyliquid64:
Please review the previous PR first - this builds off that.


exec/internal/compile/compile.go
```diff
@@ -131,16 +137,38 @@ type block struct {
...
 
+// BytecodeMetadata encapsulates metadata about a bytecode stream.
+type BytecodeMetadata struct {
+	BranchTables []*BranchTable
+	Instructions []InstructionMetadata
+
+	// Inbound jumps - used by the AOT/JIT scanner to
+	// avoid generating native code which has an inbound
+	// jump target somewhere deep inside.
+	InboundTargets map[int64]bool
```
sbinet: depending on the number of elements this map may hold, we might consider using `map[int64]struct{}` instead.

exec/internal/compile/compile.go
```diff
@@ -259,6 +294,7 @@ func Compile(disassembly []disasm.Instr) ([]byte, []*BranchTable) {
 			binary.Write(buffer, binary.LittleEndian, int64(0))
 			continue
 		case ops.BrIf:
+			emitMetadata(OpJmpNz, buffer.Len(), 18)
```
sbinet: please document what the special value `18` represents and/or how it was obtained.
perhaps consider making it a `const` like `instAndInt64Len` ?

exec/internal/compile/scanner.go
```diff
@@ -0,0 +1,114 @@
...
+
+type scanner struct {
+	supportedOpcodes map[byte]bool
+}
+
+// InstructionMetadata describes a bytecode instruction.
+type InstructionMetadata struct {
+	Op    byte
+	Start int
+	Size  int
```
sbinet: we should probably spell out in what units the `Size` is.

exec/internal/compile/scanner.go
```diff
@@ -0,0 +1,114 @@
...
+)
+
+type scanner struct {
+	supportedOpcodes map[byte]bool
+}
+
+// InstructionMetadata describes a bytecode instruction.
+type InstructionMetadata struct {
+	Op    byte
+	Start int
```
sbinet: we should probably document what `Start` represents, where it starts and what are the expected units.

exec/internal/compile/scanner.go
```diff
@@ -0,0 +1,114 @@
...
+// license that can be found in the LICENSE file.
+
+package compile
+
+import (
+	ops "github.com/go-interpreter/wagon/wasm/operators"
+)
+
+type scanner struct {
+	supportedOpcodes map[byte]bool
```
sbinet: depending on the expected size of this map, we might want to consider replacing it with `map[byte]struct{}` (that is slightly leaner in terms of memory usage.)

exec/internal/compile/scanner.go
```diff
@@ -0,0 +1,114 @@
...
+type InstructionMetadata struct {
+	Op    byte
+	Start int
+	Size  int
+}
+
+// CompilationCandidate describes a range of bytecode that can
+// be translated to native code.
+type CompilationCandidate struct {
+	// Bytecode index of the first opcode.
```
sbinet: for short doc strings like that, it looks nicer to put it on the same line than the field being documented.

exec/internal/compile/scanner.go
```diff
@@ -0,0 +1,114 @@
...
+	Beginning uint
```
sbinet: `CompilationCandidate` doesn't seem to use a consistent term for "start": `Beginning`, `StartInstruction`.
it's consistent for "end" (`End`, `EndInstruction`.)
could we fix that?

exec/internal/compile/scanner.go
```diff
@@ -0,0 +1,114 @@
...
+func (s *CompilationCandidate) reset() {
+	s.Beginning = 0
+	s.End = 0
+	s.StartInstruction = 0
+	s.EndInstruction = 0
+	s.Metrics = Metrics{}
+}
+
+// Bounds returns the beginning & end index in the bytecode which
+// this candidate would replace.
```
sbinet: we should probably document whether this is a half-open interval or not :)

exec/native_compile.go
```diff
@@ -0,0 +1,144 @@
...
+import (
+	"encoding/binary"
+	"fmt"
+	"runtime"
+
+	"github.com/go-interpreter/wagon/exec/internal/compile"
+	ops "github.com/go-interpreter/wagon/wasm/operators"
+)
+
+// Parameters which decide when a sequence should be compiled.
```
sbinet: `s/which/that/`
`s/when/whether/`

exec/vm.go
```diff
@@ -70,17 +73,39 @@ type VM struct {
...
 var endianess = binary.LittleEndian
 
-// NewVM creates a new VM from a given module. If the module defines a
-// start function, it will be executed.
-func NewVM(module *wasm.Module) (*VM, error) {
+type config struct {
+	EnableAOT bool
+}
+
+// VMOptions describes a customization that can be applied to the VM.
```
sbinet: `s/VMOptions/VMOption/`

exec/vm.go
```diff
@@ -160,6 +196,12 @@ func NewVM(module *wasm.Module) (*VM, error) {
 	return &vm, nil
 }
 
+// NewVM creates a new VM from a given module. If the module defines a
+// start function, it will be executed.
+func NewVM(module *wasm.Module) (*VM, error) {
```
sbinet: as commented on the previous incarnation of this PR, I think it's fair game to just call the function you name `NewVMWithOptions` simply `NewVM`.

exec/vm.go
```diff
@@ -326,6 +378,8 @@ outer:
 	for int(vm.ctx.pc) < len(vm.ctx.code) && !vm.abort {
 		op := vm.ctx.code[vm.ctx.pc]
 		vm.ctx.pc++
+		// fmt.Printf("stack=%+v (%d,%d), locals=%+v\n", vm.ctx.stack, len(vm.ctx.stack), cap(vm.ctx.stack), vm.ctx.locals)
```
sbinet: please remove.

exec/internal/compile/scanner.go
```diff
@@ -0,0 +1,114 @@
...
+		inProgress.Metrics.AllOps++
+	}
+
+	// End of instructions - emit the inProgress candidate if
+	// its at least 3 instructions.
+	if inProgress.Metrics.AllOps > 2 {
+		finishedCandidates = append(finishedCandidates, inProgress)
+	}
+
+	//fmt.Printf("Candidates: %+v\n", finishedCandidates)
```
sbinet: please remove.

exec/vm.go
```diff
@@ -401,6 +459,17 @@ outer:
...
 }
 
+// Close frees any resources managed by the VM.
+func (vm *VM) Close() error {
+	if vm.nativeBackend != nil {
+		if err := vm.nativeBackend.Close(); err != nil {
+			return err
+		}
+	}
+	vm.abort = true // prevents further use.
```
sbinet: is it ok to leave this to `false` if `vm.nativeBackend.Close()` errors out?

exec/native_compile.go
```diff
@@ -0,0 +1,144 @@
...
+
+	for i := range vm.funcs {
+		if _, isGoFunc := vm.funcs[i].(*goFunction); isGoFunc {
+			continue
+		}
+
+		fn := vm.funcs[i].(compiledFunction)
+		candidates, err := vm.nativeBackend.Scanner.ScanFunc(fn.code, fn.codeMeta)
+		if err != nil {
+			return fmt.Errorf("AOT scan failed on vm.funcs[%d]: %v", i, err)
```
sbinet: please prepend the package name so users can help us understand the error :)

exec/native_compile.go
```diff
@@ -0,0 +1,144 @@
...
+				continue
+			}
+			lower, upper := candidate.Bounds()
+			if (upper - lower) < minInstBytes {
+				continue
+			}
+
+			asm, err := vm.nativeBackend.Builder.Build(candidate, fn.code, fn.codeMeta)
+			if err != nil {
+				return fmt.Errorf("native compilation failed on vm.funcs[%d].code[%d:%d]: %v", i, lower, upper, err)
```
sbinet: ditto.
perhaps we should instead introduce proper error values/types?

exec/vm.go
twitchyliquid64: I guess not - some executable pages would be unmapped. I'll move it up.

exec/internal/compile/scanner.go
sbinet: ok.

exec/internal/compile/scanner.go
sbinet: sorry, maths/physics-y speak, I guess :)
I meant to say we should probably say whether the interval we are describing holds the end part or excludes it.

_ie:_ 
- `[1, 2, 3, 4] --> [2, 3]` (when `Beginning=1` and `End=3`)
- `[1, 2, 3, 4] --> [2, 3, 4]` (when `Beginning=1` and `End=3`)

usual off-by-one errors... :)

twitchyliquid64:
:smile: :airplane:

exec/internal/compile/scanner.go
twitchyliquid64: Ah, I see :)

sbinet:
please rebase on top the latest `master`. thanks.

twitchyliquid64:
Okay, I think I got that to work at the expense of the commit history (I think that can be squashed at merge?), LGTU?

exec/exec_test.go
```diff
@@ -388,7 +396,7 @@ func BenchmarkModules(b *testing.B) {
 			if err != nil {
 				b.Fatal(err)
 			}
-			runTest(path, testCases, b)
+			runTest(path, testCases, b, false)
```
sbinet: I haven't checked yet in the later PRs but I assume we'll have the option to benchmark with native=true :)

twitchyliquid64: For now, we run both the tests twice (once with native and once with non-native). If you are on a platform with no native support, both tests are identical. I can change the behaviour if you would rather only run them once using a build tag.

sbinet: the current behaviour is fine for tests.
for benchmarks, I guess it would be great to have a way to show the AOT improvements.
(but this can wait for when the PRs w/ the actual impl is there.)

twitchyliquid64: SGTM :) I plan on the following (not in PRs yet, but sometime in the future):

- Exposing an API on the VM/func so you can see what sequences were compiled to native code
- Fixing the benchmarks, making a few common cases (like hashing)
 - Writing some docs on the status of the native backend & how to enable it.

PR #113, opened 2019-03-24
3/4 - Implement native compilation - Implement amd64 backend.

twitchyliquid64:
Please review/merge the earlier PRs in this series first - this builds on that.

This PR implements the memory allocator & a very basic amd64 backend.


exec/internal/compile/allocator.go
```diff
@@ -0,0 +1,67 @@
...
+	for _, block := range a.blocks {
+		if err := block.mem.Unmap(); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// AllocateExec allocates a block of executable memory with the given code contained.
+func (a *MMapAllocator) AllocateExec(asm []byte) (NativeCodeUnit, error) {
```
sbinet: do you foresee a strategy to reclaim mmap allocated blocks?
(just asking. I understand this can be quite tricky and might require a fair amount of additional infrastructure...)

exec/internal/compile/backend_amd64.go
```diff
@@ -0,0 +1,311 @@
...
+)
+
+// dirtyRegs hold booleans that are true when the register stores
+// a reserved value that needs to be flushed to memory.
+type dirtyRegs struct {
+	R12 bool
+	R13 bool
+}
+
+// Details of the AMD64 backend:
```
sbinet: are those details inherited from the Go AMD64 backend or just how we decided to wield it?

exec/internal/compile/backend_amd64.go
```diff
@@ -0,0 +1,311 @@
...
+				return nil, fmt.Errorf("emitBinaryI64: %v", err)
+			}
+		default:
+			return nil, fmt.Errorf("cannot handle inst[%d].Op 0x%x", i, inst.Op)
+		}
+	}
+	b.emitPostamble(builder, &regs)
+
+	out := builder.Assemble()
+	// cmd := exec.Command("ndisasm", "-b64", "-")
```
sbinet: there are many of these snippets of code being commented out.
I understand they were quite useful to debug things.

perhaps we should replace them with a proper helper func `ndisasm`, guarded with either a debug flag or an environment variable?

or just remove them :)

exec/internal/compile/backend_amd64.go
```diff
@@ -0,0 +1,311 @@
...
+// Build implements exec.instructionBuilder.
+func (b *AMD64Backend) Build(candidate CompilationCandidate, code []byte, meta *BytecodeMetadata) ([]byte, error) {
+	builder, err := asm.NewBuilder("amd64", 128)
+	if err != nil {
+		return nil, err
+	}
+	var regs dirtyRegs
+	b.emitPreamble(builder, &regs)
+
+	for i := candidate.StartInstruction; i <= candidate.EndInstruction; i++ {
```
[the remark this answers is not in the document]
twitchyliquid64: I think so? `EndInstruction` is set twice and read once in the whole project, so I dont think we will break anything. I'll make the change.

exec/internal/compile/allocator.go
twitchyliquid64: We are keeping a slice of the allocated blocks. The plan was to iterate through the first x blocks to see if there is any free space we can use instead of allocating a new page. I saved this work for a future PR.

All pages are freed on a call to `MMapAllocator.Close()`, which is called on a call to `vm.Close()`.

exec/internal/compile/backend_amd64.go
twitchyliquid64: Just how I decided to wield it right now. I plan on changing this up to be more dynamic / efficient in the future - this is just a naive first implementation.

(In particular, most of the instructions emitted around handling go slices could be omitted, and intermediate values between operations could be stored in registers rather than always pushed to the wasm stack).

exec/internal/compile/backend_amd64.go
twitchyliquid64: I'm a fan of leaving them in as they are *very* useful :), and to a new contributor it may be helpful for them to see what is happening. I'll change them all to calls to `debugPrintAsm()`.

twitchyliquid64:
:grin: :tada:

exec/internal/compile/backend_amd64.go
sbinet: ok. thanks.

PR #114, opened 2019-03-24
4/4 - Implement native compilation - plug backend + more tests.

twitchyliquid64:
Please review/merge the other PRs in this series first - this builds on those commits.

This is the final PR that enables a minimal (but working) amd64 native backend.

Future PRs will be small additions to support more and more opcodes.


twitchyliquid64:
I've merged master in, this should be good for review :)

exec/native_compile_test.go
```diff
@@ -0,0 +1,189 @@
...
+			},
+		},
+		nativeBackend: nc,
+	}
+	vm.newFuncTable()
+
+	// setup mocks.
+	nc.Scanner.(*mockSequenceScanner).emit = []compile.CompilationCandidate{
+		// Sequence with single integer op - should not compiled.
+		compile.CompilationCandidate{Start: 0, End: 7, EndInstruction: 3, Metrics: compile.Metrics{IntegerOps: 1}},
```
sbinet: looking at the history of the changes, it doesn't seem to me `EndInstruction`'s value was changed here, nor below.
shouldn't they be (resp.) `4` and `10` ?

twitchyliquid64: Fixed.

Note: Start/EndInstruction is not part of the code under test, this test just tests `tryNativeCompile()`, and mocks everything else.

PR #117, opened 2019-04-02
Implement XOR / SHR / SHL on amd64 native backend.

twitchyliquid64:
As before, relies on previous PRs.


exec/internal/compile/backend_amd64_test.go
```diff
@@ -239,12 +269,16 @@ func TestAMD64OperationsI64(t *testing.T) {
...
 				t.Fatal(err)
 			}
-			b.emitPreamble(builder, regs)
 
+			b.emitPreamble(builder, regs)
 			for _, arg := range tc.Args {
 				b.emitPushI64(builder, regs, arg)
 			}
-			b.emitBinaryI64(builder, regs, tc.Op)
+			if tc.Op == ops.I64Shl || tc.Op == ops.I64ShrU || tc.Op == ops.I64ShrS {
```
sbinet: I'd rather use a `switch` here:
```go
switch tc.Op {
case ops.I64Shl, ops.I64ShrU, ops.I64ShrS:
    ...
default:
    ...
}
```

twitchyliquid64: Done :)

twitchyliquid64:
:airplane: :)

PR #119, opened 2019-04-05
Implement comparison Ops, fix bug with branch table target tracking, add benchmark.

twitchyliquid64:
- Implements greater-than, less-than etc ops
 - Fixed a bug where branch table jump addresses were not being tracked, causing the scanner to emit invalid candidates
 - Created an ideal-case benchmark for Native vs Interpreted execution modes.

On my i7 2nd gen:

```
[xxx@xxx]:~/projects/wagon> go test -bench=BenchmarkU64Arithmetic github.com/go-interpreter/wagon/exec
goos: linux
goarch: amd64
BenchmarkU64Arithmetic10Interpreted-8   	  100000	     16611 ns/op
BenchmarkU64Arithmetic10Native-8        	  200000	      5781 ns/op
BenchmarkU64Arithmetic50Interpreted-8   	   20000	     76431 ns/op
BenchmarkU64Arithmetic50Native-8        	   50000	     25961 ns/op
PASS
ok  	github.com/go-interpreter/wagon/exec	7.565s
```

So currently 3x, but heaps of room for improvement (I think I can cut emitted assembly by 80%, and theres still heaps of opcodes we dont yet support). I'm expecting we should be able to hit 10-20x here.


exec/exec_test.go
```diff
@@ -408,3 +409,61 @@ func TestNonSpec(t *testing.T) {
...
+	vm, err := exec.NewVM(module, exec.EnableAOT(nativeBackend))
+	if err != nil {
+		t.Fatalf("%s: %v", fileName, err)
+	}
+	return vm, module.Export.Entries[funcName].Index
+}
+
+func BenchmarkU64Arithmetic10Interpreted(b *testing.B) {
+	vm, funcIndex := loadModuleFindFunc(b, "testdata/rust-basic.wasm", "loopedArithmeticI64Benchmark", false)
+	for n := 0; n < b.N; n++ {
```
sbinet: perhaps insert a `b.ResetTimer()` before the for-loop?

exec/exec_test.go
```diff
@@ -408,3 +409,61 @@ func TestNonSpec(t *testing.T) {
...
+	}
+}
+
+func BenchmarkU64Arithmetic10Native(b *testing.B) {
+	if runtime.GOARCH != "amd64" {
+		b.SkipNow()
+	}
+
+	vm, funcIndex := loadModuleFindFunc(b, "testdata/rust-basic.wasm", "loopedArithmeticI64Benchmark", true)
+	for n := 0; n < b.N; n++ {
```
sbinet: perhaps insert a `b.ResetTimer()` before the for-loop?

exec/exec_test.go
```diff
@@ -408,3 +409,61 @@ func TestNonSpec(t *testing.T) {
...
+		vm.ExecCode(int64(funcIndex), 10, 10)
+	}
+}
+
+func BenchmarkU64Arithmetic50Interpreted(b *testing.B) {
+	vm, funcIndex := loadModuleFindFunc(b, "testdata/rust-basic.wasm", "loopedArithmeticI64Benchmark", false)
+	for n := 0; n < b.N; n++ {
```
sbinet: perhaps insert a `b.ResetTimer()` before the for-loop?

exec/exec_test.go
```diff
@@ -408,3 +409,61 @@ func TestNonSpec(t *testing.T) {
...
+	}
+}
+
+func BenchmarkU64Arithmetic50Native(b *testing.B) {
+	if runtime.GOARCH != "amd64" {
+		b.SkipNow()
+	}
+
+	vm, funcIndex := loadModuleFindFunc(b, "testdata/rust-basic.wasm", "loopedArithmeticI64Benchmark", true)
+	for n := 0; n < b.N; n++ {
```
sbinet: perhaps insert a `b.ResetTimer()` before the for-loop?

exec/exec_test.go
```diff
@@ -408,3 +409,61 @@ func TestNonSpec(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("%s: %v", fileName, err)
+	}
+	return vm, module.Export.Entries[funcName].Index
+}
+
+func BenchmarkU64Arithmetic10Interpreted(b *testing.B) {
+	vm, funcIndex := loadModuleFindFunc(b, "testdata/rust-basic.wasm", "loopedArithmeticI64Benchmark", false)
+	for n := 0; n < b.N; n++ {
+		vm.ExecCode(int64(funcIndex), 10, 10)
```
sbinet: one should probably store one of the 2 returned values into a global variable to prevent unwanted optimization from the compiler.

exec/exec_test.go
```diff
@@ -408,3 +409,61 @@ func TestNonSpec(t *testing.T) {
...
+}
+
+func BenchmarkU64Arithmetic10Native(b *testing.B) {
+	if runtime.GOARCH != "amd64" {
+		b.SkipNow()
+	}
+
+	vm, funcIndex := loadModuleFindFunc(b, "testdata/rust-basic.wasm", "loopedArithmeticI64Benchmark", true)
+	for n := 0; n < b.N; n++ {
+		vm.ExecCode(int64(funcIndex), 10, 10)
```
sbinet: ditto (and below as well.)

exec/internal/compile/scanner.go
```diff
@@ -94,7 +94,16 @@ func (s *scanner) ScanFunc(bytecode []byte, meta *BytecodeMetadata) ([]Compilati
...
 			inProgress.Metrics.StackWrites++
+		case ops.SetLocal:
+			inProgress.Metrics.IntegerOps++
+			inProgress.Metrics.StackReads++
+		case ops.I64Eqz:
+			inProgress.Metrics.IntegerOps++
+			inProgress.Metrics.StackReads++
+			inProgress.Metrics.StackWrites++
 
+		case ops.I64Eq, ops.I64Ne, ops.I64LtU, ops.I64GtU, ops.I64LeU, ops.I64GeU:
```
sbinet: unless we expect additional operations to be carried for during the case, I would just merge all these cases together, possibly leaving each of these different cases sets regrouped on their own line:
```go
case ops.I64Eq, ops.I64Ne, ops.I64LtU, ops.I64GtU, ops.I64LeU, ops.I64GeU,
    ops.I64Shl, ops.I64ShrU, ops.I64ShrS,
    ops.I64DivU, ops.I64RemU, ops.I64DivS, ops.I64RemS,
    ops.I64Add, ops.I64Sub, ops.I64Mul, ops.I64And, ops.I64Or, ops.I64Xor:
```

apologies for missing this earlier :)

exec/internal/compile/compile.go
```diff
@@ -276,7 +276,7 @@ func Compile(disassembly []disasm.Instr) ([]byte, *BytecodeMetadata) {
 			}
 
 			for _, table := range block.branchTables {
-				table.patchTable(table.blocksLen-depth-1, int64(block.offset))
+				table.patchTable(table.blocksLen-depth-1, int64(block.offset), inboundTargets)
```
sbinet: do we have an explicit test that exercizes this?

twitchyliquid64: Indirectly :) `TestSpec/testdata/spec/switch.wasm` has br_tables for dayz, and now that more instructions are supported (and hence more compilation candidates), the compiler was emitting code for instructions that the branch table was jumping into halfway. This change fixes that.

exec/internal/compile/scanner.go
twitchyliquid64: Ack, will do in morning :)

exec/exec_test.go
twitchyliquid64: Ack, will do in morning :)

exec/exec_test.go
twitchyliquid64: Ack, will do in morning :)

twitchyliquid64:
:airplane: :unicorn:

PR #120, opened 2019-04-07
amd64 compiler optimizations: Improves speed by 30%.

twitchyliquid64:
- Pairs of instructions that can be re-written into a single amd64 instruction (ie: basic operations that accept an immediate value as an operand) are rewritten into that form. For example, a `i64.Const` + `i64.Add` produce a single instruction `ADDQ <register>, <immediate value>`.

 - Opportunistically keep track of the size of the stack in a register rather than always reading from memory (The value is flushed if necessary in the postamble).

 - Opportunistically cache a pointer to the stack and/or local backing array in a register rather than always reading for memory. As this value is immutable we don't need to flush it back in the postamble.


exec/internal/compile/backend_amd64.go
```diff
@@ -14,21 +14,61 @@ import (
...
+	R15 dirtyState
+}
+
+func (regs *dirtyRegs) flush(builder *asm.Builder, reg uint16) {
+	var regState *dirtyState
+	switch reg {
+	case x86.REG_R13:
+		regState = &regs.R13
+	default:
+		panic(fmt.Sprintf("unknown register: %v", reg))
```
sbinet: please prepend the package name to the panic string.

exec/internal/compile/backend_amd64.go
```diff
@@ -14,21 +14,61 @@ import (
...
+	case stateStackLen:
+		prog := builder.NewProg()
+		prog.As = x86.AMOVQ
+		prog.From.Type = obj.TYPE_REG
+		prog.From.Reg = x86.REG_R13
+		prog.To.Type = obj.TYPE_MEM
+		prog.To.Reg = x86.REG_R10
+		prog.To.Offset = 8
+		builder.AddInstruction(prog)
+	}
```
sbinet: perhaps add a `default` case that panics? just to be sure?

exec/internal/compile/backend_amd64.go
```diff
@@ -14,21 +14,61 @@ import (
...
 }
 
 // Details of the AMD64 backend:
 // Reserved registers (for now):
 //  - R10 - pointer to stack sliceHeader
 //  - R11 - pointer to locals sliceHeader
-//  - R12 - pointer for stack item
+//  - R12 - reserved for stack handling
 //  - R13 - stack size
+// Puesdo-scratch registers (can be used for scratch as long as their
```
sbinet: `s/Puesdo/Pseudo/`

exec/internal/compile/backend_amd64.go
```diff
@@ -88,6 +128,32 @@ func (b *AMD64Backend) Build(candidate CompilationCandidate, code []byte, meta *
 	for i := candidate.StartInstruction; i < candidate.EndInstruction; i++ {
 		//fmt.Printf("i=%d, meta=%+v, len=%d\n", i, meta.Instructions[i], len(code))
 		inst := meta.Instructions[i]
+
+		// Optimization: Const followed by binary instruction: sometimes can be
+		// reduced to a single operation.
+		if (inst.Op == ops.I64Const || inst.Op == ops.I32Const) && (i+1) < candidate.EndInstruction {
+			imm := b.readIntImmediate(code, inst)
+			nextInst := meta.Instructions[i+1]
+			if _, ok := rhsConstOptimizable[nextInst.Op]; ok && imm < 256 && imm >= 0 {
```
sbinet: I would use a `switch` here:
```go
switch _, ok := rhsConstOptimizable[nextInst.Op]; {
case ok && 0 <= imm && imm < 256:
    ...
case nextInst.Op == ops.SetLocal:
    ...
}
```
(and rewrite the condition on `imm` to read more like an interval: `0 <= imm && imm < 256`)

exec/internal/compile/backend_amd64.go
```diff
@@ -365,6 +432,33 @@ func (b *AMD64Backend) emitBinaryI64(builder *asm.Builder, regs *dirtyRegs, op b
...
+	case ops.I64Add:
+		prog.As = x86.AADDQ
+	case ops.I64Sub:
+		prog.As = x86.ASUBQ
+	case ops.I64Shl:
+		prog.As = x86.ASHLQ
+	case ops.I64ShrU:
+		prog.As = x86.ASHRQ
+	default:
+		return fmt.Errorf("cannot handle op: %x", op)
```
sbinet: please prepend the package name.

sbinet:
ah, also: could you put the before/after speed improvements in one of the commit messages?
(probably using the benchmarks and [golang.org/x/perf/cmd/benchstat](https://godoc.org/golang.org/x/perf/cmd/benchstat))

twitchyliquid64:
I've put the benchmark in the commit message, but I will note benchmarks will vary alot depending on the cache size and generation of processor (especially here, where we are emitting raw assembly).

For instance, the interpreted benchmark is roughly the same between my laptop (8th gen, 8mb cache) and desktop (3rd gen, 6mb cache), but the native execution benchmarks are all ~15% faster. This is [largely attributable](https://www.agner.org/optimize/microarchitecture.pdf) to a better uop cache, instruction fusion, larger instruction-fetch blocks, larger reorder buffer etc. These difference will get larger the further we move from reading/writing to memory (as is with the current approach, where intermediate values are written into the stack slice) to storing intermediates in registers.

exec/internal/compile/backend_amd64.go
twitchyliquid64: Nack: this is called from `AMD64Backend.Build()`, which will rewrite the error like so: ` fmt.Errorf("compile: amd64.emitRHSConstOptimizedInstruction: %v", err)`.

sbinet: fair enough.

PR #128, opened 2019-06-07
Add benchmarks for f32 & f64 operations.

twitchyliquid64:
```
[xxx@xxx]:~/projects/wagon/exec> go test -bench=BenchmarkF github.com/go-interpreter/wagon/exec
goos: linux
goarch: amd64
BenchmarkF64Arithmetic10Interpreted-8   	  100000	     14344 ns/op
BenchmarkF64Arithmetic10Native-8        	  300000	      4343 ns/op
BenchmarkF32Arithmetic10Interpreted-8   	  100000	     13365 ns/op
BenchmarkF32Arithmetic10Native-8        	  300000	      4240 ns/op
PASS
ok  	github.com/go-interpreter/wagon/exec	7.688s
```

Looks like the floating point operations are only ~3x faster instead of ~5x. I think this is attributable to missing operations for conversion from integer to float - hopefully I'll get around to that this weekend.


exec/exec_test.go
```diff
@@ -465,6 +465,46 @@ func BenchmarkU64Arithmetic10Native(b *testing.B) {
...
+		benchmarkDummy, _ = vm.ExecCode(int64(funcIndex), 10, 10)
+	}
+}
+
+func BenchmarkF32Arithmetic10Native(b *testing.B) {
+	if runtime.GOARCH != "amd64" {
+		b.SkipNow()
+	}
+
+	vm, funcIndex := loadModuleFindFunc(b, "testdata/rust-basic.wasm", "loopedArithmeticF64Benchmark", true)
```
sbinet: `s/F64/F32/` ?

twitchyliquid64: ^^'

sbinet:
this should have "improved" the benchmarks, right? :)

twitchyliquid64:
Dont think so, as processors arent much faster on 32bit floats than 64bit floats?

sbinet:
my naive assumption was that pipelines would be somewhat more full.

PR #129, opened 2019-06-07
Implement global load/set, memory 64bit/32bit load

twitchyliquid64:
This PR starts implementing more tablestakes.

All the complexity here is in mitigating Spectre. We use the same technique as v8, which is this:

 * Set a register to `0xffff`... before the bounds check
 * Zero the register in a non-branching way if out-of-bounds
 * Mask the result read from the the memory array with the mask register

This works because CPUs do not speculate through bitmask operations, and because if the bounds check was mis-speculated, the poison register would be set to zero. As a result, attacker-controlled instructions later on would always get a zero value for their speculation attack, so they cannot build any measurement primitives to determine the value from the speculated out-of-bounds access.


twitchyliquid64:
With these changes (and the last few PRs), the integer benchmark is now 50% faster than when it was [introduced](https://github.com/go-interpreter/wagon/pull/119).

```
[xxx@xxx]:~/projects/wagon> go test -bench=BenchmarkU github.com/go-interpreter/wagon/exec
goos: linux
goarch: amd64
BenchmarkU64Arithmetic10Interpreted-8   	  100000	     17064 ns/op
BenchmarkU64Arithmetic10Native-8        	  300000	      3787 ns/op
BenchmarkU64Arithmetic50Interpreted-8   	   20000	     76593 ns/op
BenchmarkU64Arithmetic50Native-8        	  100000	     16423 ns/op
PASS
ok  	github.com/go-interpreter/wagon/exec	9.162s
```

PR #131, opened 2019-06-20
wasm: when function names are present, include them

justinclift:
LLVM generates (optional) function names, including them in a custom section on the end of the .wasm file.

It's useful to have them handy when working with decoded wasm data, so adding them directly to the Function struct seems like a good spot.

---

This PR originally included code by @mastersingh24, from the Life project.  The below comment was regarding allowing that piece of code to be incorporated here with BSD licencing.

However, it turns out to be unneeded, as Wagon already has the ability to decode Custom Name sections, so it was simpler to reuse that.


justinclift:
Hmmm.  Investigating the test failure now.

justinclift:
Heh Heh Heh.  Looks like Life actually took the code from some other part of Wagon:

https://github.com/perlin-network/life/commit/3f69e3d0fadecf38aec537869f48f62d2f798481#diff-fe9f64fa1f23ebc8ef6646f2637f2283

justinclift:
@mastersingh24 This PR adds reading custom name section data, based on your commit in the Life repo.

However, these repos have different licences (BSD vs MIT).  Are you ok with this code becoming BSD licenced in this instance, for this purpose?

mastersingh24:
@justinclift  - I don't have any issues with that

justinclift:
@mastersingh24 Excellent.  Thanks. :smile:

justinclift:
As an aside, the `custom_sections.wasm` file in the test data seems a bit strange:

```
$ wasm2wat custom_section.wasm 
0000058: error: invalid function count 36, only 1 bytes left in section
```

justinclift:
Looking into adding some new test data now, incorporating a custom section with function names.

justinclift:
Adding some TinyGo (via LLVM) generated test data with function names in custom sections probably isn't going to be workable at the moment. :frowning:

Testing various TinyGo generated .wasm files locally here, and the way encoding the `0` value is done for Call instructions is strange.

They're padded with four `0x80` bytes prior to the `0x0`, which while technically valid, means they're not binary identical when Wagon reassembles them (more efficiently) for the `asm_test.go` test:

https://github.com/go-interpreter/wagon/blob/c2df2531f44d5cf328e9abce8162e71367592ea9/disasm/asm_test.go#L54-L56

justinclift:
It looks like this padding is common to all .wasm generated via LLVM at the moment.  At least with LLVM ~8.0.1-rc2.

Saying that after trying with C compiled to wasm too (using `clang`).  Same problem.

Using `wasm-ld` with `--compress-relocations` does seem to remove the padding bytes (when quickly checking via `hexedit`), however that option requires either `--strip-debug` or `--strip-all`, both of which remove the function names from the end of the file. :man_shrugging:

justinclift:
Just created a PR adding @mastersingh24 and myself to the CONTRIBUTORS and AUTHORS files, as per the contributor guidelines: https://github.com/go-interpreter/license/pull/16

justinclift:
Actually, hold off on this for the moment.  Just realised it's not filling in the name field for imported functions.  The info is available, so should be incorporated.  Will get that added. :smile:

justinclift:
k, should be good now. :smile:

justinclift:
Hmmm, looks like Wagon already has support for parsing function names from custom name section data: https://github.com/go-interpreter/wagon/blob/c2df2531f44d5cf328e9abce8162e71367592ea9/wasm/section.go#L1002-L1078

justinclift:
Just created this, using the exist Wagon unmarshalling code for name sections.

Sorry @mastersingh24... your code wasn't needed after all. :wink:

justinclift:
Not 100% sure how at the moment.  No LLVM generated wasm will pass the current Wagon test suite due to the Call function padding problem (above). :frowning:

If I can somehow get something generated which incorporates custom functions, then I can write a test.

justinclift:
Oh.  I'll investigate using Wagon to generate .wasm files directly.  That might work (haven't looked). :smile:

justinclift:
Killed the branch when cleaning up my fork, and it turns out GitHub uses that as a signal to automatically close the PR.  Oops. :wink:

sbinet:
to follow-up on this, one possible avenue (instead of relying on the LLVM machinery), one could perhaps generate a `.wasm` file from this definition:
```
(module
  (type (;0;) (func (param i32 i32) (result i32)))
  (func $add (type 0) (param i32 i32) (result i32)
    get_local 0
    get_local 1
    i32.add)
  (export "add" (func $add))
)
```

and make sure we correctly parse that `.wasm` file and display `add` wherever expected.

WDYT?

wasm/index.go
```diff
@@ -38,14 +39,43 @@ func (m *Module) populateFunctions() error {
 		return nil
 	}
 
+	// If present, extract the function names from the custom 'name' section (LLVM generates these)
```
agnivade: Even `gc` generates these. So I don't think there is a need to say "(LLVM generates these)".

btw, great work. Waiting for this to be merged :smile:

justinclift: Ahhh, sorry, got busy with other stuff.  I'll get back to this in the next few days. :roll_eyes:

justinclift: With the "LLVM generates these", no worries, can easily drop it.  Just been doing stuff pretty exclusively with LLVM generated wasm recently, so haven't checked otherwise. :wink:

justinclift:
Ahhh, sorry, got busy with other stuff.  I'll get back to this in the next few days. :roll_eyes:

wasm/index.go
justinclift: Just removed that text fragment. :smile:

justinclift:
Well, this is interesting.  That failure is looking suspiciously like a problem with Wagon's `EncodeModule()` function not being deterministic.

Saying that because I took a fairly cheating approach to generate the wasm test data... :wink:

Wrote [a small util](https://github.com/justinclift/wasm-wagon-re-encode/blob/master/main.go) to take read in an LLVM generated .wasm file, re-encode it with Wagon, and write that out.

It looks like the encoded byte stream created by Wagon will have the exports in any order:

```
$ hexdump -C hello-world-tinygo.wasm > hello-world-tinygo.wasm-dump
$ hexdump -C hello-world-tinygo.wasm-re-encoded > hello-world-tinygo.wasm-re-encoded-dump
$ diff hello-world-tinygo.wasm-dump hello-world-tinygo.wasm-re-encoded-dump
10,12c10,12
< 00000090  00 0b 5f 5f 68 65 61 70  5f 62 61 73 65 03 01 0a  |..__heap_base...|
< 000000a0  5f 5f 64 61 74 61 5f 65  6e 64 03 02 11 5f 5f 77  |__data_end...__w|
< 000000b0  61 73 6d 5f 63 61 6c 6c  5f 63 74 6f 72 73 00 02  |asm_call_ctors..|
---
> 00000090  00 0b 5f 5f 68 65 61 70  5f 62 61 73 65 03 01 11  |..__heap_base...|
> 000000a0  5f 5f 77 61 73 6d 5f 63  61 6c 6c 5f 63 74 6f 72  |__wasm_call_ctor|
> 000000b0  73 00 02 0a 5f 5f 64 61  74 61 5f 65 6e 64 03 02  |s...__data_end..|
```

Running it multiple times, even on its own reencoded output, can ~randomly change around the ordering.

That's probably going to break / disallow any non-trivial example data.

sbinet:
We're probably iterating over a map somewhere and not ensuring some order...

justinclift:
Yeah, it's a pretty classic case of that.  I'll try and take a look in the next few days, to see if changing to a slice or something with stable sorting is simple.

justinclift:
This is probably the culprit: https://github.com/go-interpreter/wagon/blob/2691113c6eae68a2402e61eb67b6cae041759f59/wasm/section.go#L579

justinclift:
Dammit, after all that, it **still** fails.  Looks like the `Assemble()` function generates different byte code again.

justinclift:
Updated that small util to re-write each of the code body sections as well, which when combined with the stable sort order should be a complete fix. :man_dancing: 

Nicely, because Wagon's byte code assembler is a bit more efficient than the LLVM one, a side effect is slightly reduced file size.

In effect, it could probably be used as a general purpose util to shave off a few more bytes from the file size. :wink:

justinclift:
Good stuff @sbinet. :smile:

Paging @agnivade.  You're up. :grinning:

agnivade:
Roger that.

PR #133, opened 2019-06-28
amd64 backend optimizations: Fused load/store, more RHS opcodes, peephole optimizer.

twitchyliquid64:
We are starting to hit the limits of a template JIT. Next PR I think I will refactor the backend into separate phases, and try and lower the stack read/writes in a later phase (so we can do register allocation across opcodes).


exec/internal/compile/backend_amd64.go
```diff
@@ -227,17 +235,16 @@ func (b *AMD64Backend) Build(candidate CompilationCandidate, code []byte, meta *
...
-				continue
+			default:
+				switch nextInst.Op {
+				case ops.SetLocal, ops.SetGlobal, ops.I64Store, ops.I32Store, ops.F64Store, ops.F32Store:
+					if err := b.emitFusedConstStore(builder, &regs, code, nextInst, ci, nextCI, imm); err != nil {
+						return nil, fmt.Errorf("compile: amd64.emitFusedConstStore: %v", err)
+					}
+					i++
+					continue
+				default:
```
sbinet: probably not needed, is it?

exec/internal/compile/optimizer_amd64.go
```diff
@@ -0,0 +1,35 @@
+package compile
```
sbinet: please add the standard license header. (with 2019 as the year)

exec/internal/compile/optimizer_amd64.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"math"
+
+	asm "github.com/twitchyliquid64/golang-asm"
+	"github.com/twitchyliquid64/golang-asm/obj"
+	"github.com/twitchyliquid64/golang-asm/obj/x86"
+)
+
+func peepholeOptimizeAMD64(builder *asm.Builder) error {
```
sbinet: would you have a simple idea on how to directly test this function?
would it be relatively easy?

exec/internal/compile/backend_amd64.go
```diff
@@ -1171,10 +1202,26 @@ func (b *AMD64Backend) emitRHSConstOptimizedInstruction(builder *asm.Builder, re
 		prog.As = x86.AADDQ
 	case ops.I64Sub:
 		prog.As = x86.ASUBQ
+	case ops.I32Add:
```
sbinet: these new ops don't seem to be tested either.

exec/internal/compile/backend_amd64.go
```diff
@@ -344,6 +356,28 @@ func (b *AMD64Backend) paramsForMemoryOp(op byte) (size uint, inst obj.As) {
 	panic("unreachable")
 }
 
+func (b *AMD64Backend) emitFusedConstStore(builder *asm.Builder, regs *dirtyRegs, code []byte, nextInst InstructionMetadata, ci, nextCI currentInstruction, imm uint64) error {
```
sbinet: this new function doesn't seem to be tested either.

exec/internal/compile/optimizer_amd64.go
twitchyliquid64: Its possible, but it wouldnt be easy as instructions are stored as a linked list and id have to build the input list and then test the output one via some messy iterator.

exec/internal/compile/backend_amd64.go
twitchyliquid64: Added test.

exec/internal/compile/backend_amd64.go
twitchyliquid64: added test.

exec/internal/compile/optimizer_amd64.go
sbinet: ok

PR #138, opened 2019-07-06
Enable use of amd64 backend on Windows & OSX.

twitchyliquid64:
This accidentally builds off an earlier PR, so its dependent on that before it can go in. Have a look at the latest commit to see the changes I intended.

This commit just enables native compilation on windows and OSX. The only platform-specific operation was creating an executable page, and ive confirmed the downstream library we were using (mmap-go) works on those two platforms.


sbinet:
was this rebased off #137?
(it seems there are a few redundant changes here...)

(apologies for the belated answer/review: off-the-grid-holidays)

twitchyliquid64:
I've merged in upstream/master and that seems to have put it back to normal.

PR #148, opened 2019-07-30
wasm: fix arbitrary memory allocation attack

### LINKED ISSUE #147: Denial of Service via Arbitrary Allocation
There are many places containing codes pattern: read length then allocate array or map, like:
```golang 
count, err := leb128.ReadVarUint32(r)
if err != nil {
	return err
}
s.Bodies = make([]FunctionBody, count)
```
So the attacker can construct a invalid wasm file, set the `count` to 2^32-1, causing the Wagon interpreter to allocate more memory than is available and irrecoverably crash the entire process.
We may need to change it to something like:
```golang
initialCap := min(count, maxInitialCap)
s.Bodies = make([]FunctionBody, 0, initialCap)
....
s.Bodies = append(s.Bodies, body)
```

laizy:
fix #147


laizy:
@sbinet, review?

wasm/read.go
```diff
@@ -5,20 +5,34 @@
...
 	}
+	return count
+}
 
-	return bytes, nil
+func readBytes(r io.Reader, n int) ([]byte, error) {
+	if n == 0 {
+		return nil, nil
+	}
+	limited := io.LimitReader(r, int64(n))
```
sbinet: I don't understand what's the advantage of using an `io.LimitedReader` + a `bytes.Buffer` over a simple `[]byte` and `io.ReadFull`.
could you explain this change?

wasm/section.go
```diff
@@ -132,7 +132,8 @@ func (m *Module) readSection(r *readpos.ReadPos) (bool, error) {
 	s.Start = r.CurPos
 
 	sectionBytes := new(bytes.Buffer)
-	sectionBytes.Grow(int(payloadDataLen))
+
+	sectionBytes.Grow(int(getInitialCap(payloadDataLen)))
```
sbinet: wouldn't it be clearer to return an error if `payloadDataLen` is larger than some max-value?
_ie:_ after line 128 above:
```go
if payloadDataLen > maxInitialCap {
    return errMaxAlloc // or something
}
```
```

File: wasm/read.go (hunk 1 above)
Comment (laizy) [reply wasm/read.go]: `n` is controlled by untrusted source. The attacker can actually send you 10 bytes, but set `n` to 1GB. so directly allocate `n` bytes memory is not safe. Use `buf.ReadFrom`, it will read the data by chunks, enlarge the internal buf on demand. In this way, we will avoid this attack, since when we read out 10 bytes, it will hit EOF.

Comment (sbinet) [reply wasm/read.go]: yes, this I understood :)
but I was under the assumption we'd always pass `n` through `getInitialCap` (or some other sanity check) before passing it to `readBytes`.

Comment (laizy) [reply wasm/read.go]: `getInitialCap` is just a hint function to avoid some internal reallocation, so I just name it `inital`, not `max`. It does not impose any constraints on `n`.

File: wasm/section.go (hunk 2 above)
Comment (laizy) [reply wasm/section.go]: unfortunately, the wasm spec does not limit that data length. And it is hard to choose a reasonable constant max value. 

- If we set it too large, it still will cause memory risk, if the attacker can send a lot of wasm files concurrently;
- If we set it too small, it may limit some use cases;
- If we left it as param to let user fill in, it is a little cumbersome;

but this pr can handle it cleanly.

File: wasm/read.go (hunk 1 above)
Comment (laizy) [reply wasm/read.go]: actually, `make([]Value, 0, getInitialCap(count))` can just be replaced with `make([]Value, 0)` if we do not care about that optimization.

File: wasm/section.go [hunk 3]
```diff
@@ -268,11 +269,14 @@ func (s *SectionTypes) ReadPayload(r io.Reader) error {
 	if err != nil {
 		return err
 	}
-	s.Entries = make([]FunctionSig, int(count))
-	for i := range s.Entries {
-		if err = s.Entries[i].UnmarshalWASM(r); err != nil {
+
+	s.Entries = make([]FunctionSig, 0, getInitialCap(count))
+	for i := uint32(0); i < count; i++ {
+		sig := FunctionSig{}
```
sbinet: we prefer `var v T` to spell out the zero value.
```suggestion
		var sig FunctionSig
```

wasm/section.go
```diff
@@ -307,12 +311,14 @@ func (s *SectionImports) ReadPayload(r io.Reader) error {
...
 		return err
 	}
-	s.Entries = make([]ImportEntry, count)
-	for i := range s.Entries {
-		err = s.Entries[i].UnmarshalWASM(r)
-		if err != nil {
+
+	s.Entries = make([]ImportEntry, 0, getInitialCap(count))
+	for i := uint32(0); i < count; i++ {
+		entry := ImportEntry{}
```
sbinet:
```suggestion
		var entry ImportEntry
```

wasm/section.go
```diff
@@ -451,12 +457,14 @@ func (s *SectionTables) ReadPayload(r io.Reader) error {
...
 		return err
 	}
-	s.Entries = make([]Table, count)
-	for i := range s.Entries {
-		err = s.Entries[i].UnmarshalWASM(r)
-		if err != nil {
+
+	s.Entries = make([]Table, 0, getInitialCap(count))
+	for i := 0; i < int(count); i++ {
+		entry := Table{}
```
sbinet:
```suggestion
		var entry Table
```

wasm/section.go
```diff
@@ -413,13 +419,13 @@ func (s *SectionFunctions) ReadPayload(r io.Reader) error {
 	if err != nil {
 		return err
 	}
-	s.Types = make([]uint32, count)
-	for i := range s.Types {
+	s.Types = make([]uint32, 0, getInitialCap(count))
+	for i := 0; i < int(count); i++ {
```
sbinet:
```suggestion
	for i := uint32(0); i < count; i++ {
```

(on 32b architectures, `int(count)` might wrap around for big `count` values)

wasm/section.go
```diff
@@ -451,12 +457,14 @@ func (s *SectionTables) ReadPayload(r io.Reader) error {
 	if err != nil {
 		return err
 	}
-	s.Entries = make([]Table, count)
-	for i := range s.Entries {
-		err = s.Entries[i].UnmarshalWASM(r)
-		if err != nil {
+
+	s.Entries = make([]Table, 0, getInitialCap(count))
+	for i := 0; i < int(count); i++ {
```
sbinet:
```suggestion
	for i := uint32(0); i < count; i++ {
```

wasm/section.go
```diff
@@ -488,12 +496,13 @@ func (s *SectionMemories) ReadPayload(r io.Reader) error {
 	if err != nil {
 		return err
 	}
-	s.Entries = make([]Memory, count)
-	for i := range s.Entries {
-		err = s.Entries[i].UnmarshalWASM(r)
-		if err != nil {
+	s.Entries = make([]Memory, 0, getInitialCap(count))
+	for i := 0; i < int(count); i++ {
```
sbinet:
```suggestion
	for i := uint32(0); i < count; i++ {
```

wasm/section.go
```diff
@@ -488,12 +496,13 @@ func (s *SectionMemories) ReadPayload(r io.Reader) error {
...
+		entry := Memory{}
```
sbinet:
```suggestion
		var entry Memory
```

wasm/section.go
```diff
@@ -525,13 +534,15 @@ func (s *SectionGlobals) ReadPayload(r io.Reader) error {
...
 		return err
 	}
-	s.Globals = make([]GlobalEntry, count)
+
+	s.Globals = make([]GlobalEntry, 0, getInitialCap(count))
 	logger.Printf("%d global entries\n", count)
-	for i := range s.Globals {
-		err = s.Globals[i].UnmarshalWASM(r)
-		if err != nil {
+	for i := 0; i < int(count); i++ {
```
sbinet:
```suggestion
	for i := uint32(0); i < count; i++ {
```

wasm/section.go
```diff
@@ -525,13 +534,15 @@ func (s *SectionGlobals) ReadPayload(r io.Reader) error {
...
+		global := GlobalEntry{}
```
sbinet:
```suggestion
		var global GlobalEntry
```

wasm/section.go
```diff
@@ -708,12 +720,14 @@ func (s *SectionElements) ReadPayload(r io.Reader) error {
 	if err != nil {
 		return err
 	}
-	s.Entries = make([]ElementSegment, count)
-	for i := range s.Entries {
-		err = s.Entries[i].UnmarshalWASM(r)
-		if err != nil {
+
+	s.Entries = make([]ElementSegment, 0, getInitialCap(count))
+	for i := 0; i < int(count); i++ {
```
sbinet:
```suggestion
	for i := uint32(0); i < count; i++ {
```

wasm/section.go
```diff
@@ -708,12 +720,14 @@ func (s *SectionElements) ReadPayload(r io.Reader) error {
...
+		element := ElementSegment{}
```
sbinet:
```suggestion
		var element ElementSegment
```

wasm/section.go
```diff
@@ -751,14 +765,13 @@ func (s *ElementSegment) UnmarshalWASM(r io.Reader) error {
 	if err != nil {
 		return err
 	}
-	s.Elems = make([]uint32, numElems)
-
-	for i := range s.Elems {
+	s.Elems = make([]uint32, 0, getInitialCap(numElems))
+	for i := 0; i < int(numElems); i++ {
```
sbinet:
```suggestion
	for i := uint32(0); i < numElems; i++ {
```

wasm/section.go
```diff
@@ -798,14 +811,16 @@ func (s *SectionCode) ReadPayload(r io.Reader) error {
 	if err != nil {
 		return err
 	}
-	s.Bodies = make([]FunctionBody, count)
+	s.Bodies = make([]FunctionBody, 0, getInitialCap(count))
 	logger.Printf("%d function bodies\n", count)
 
-	for i := range s.Bodies {
+	for i := 0; i < int(count); i++ {
```
sbinet:
```suggestion
	for i := uint32(0); i < count; i++ {
```

wasm/section.go
```diff
@@ -798,14 +811,16 @@ func (s *SectionCode) ReadPayload(r io.Reader) error {
...
 		logger.Printf("Reading function %d\n", i)
-		if err = s.Bodies[i].UnmarshalWASM(r); err != nil {
+		body := FunctionBody{}
```
sbinet:
```suggestion
		var body FunctionBody
```

wasm/section.go
```diff
@@ -849,12 +863,14 @@ func (f *FunctionBody) UnmarshalWASM(r io.Reader) error {
 	if err != nil {
 		return err
 	}
-	f.Locals = make([]LocalEntry, localCount)
+	f.Locals = make([]LocalEntry, 0, getInitialCap(localCount))
 
-	for i := range f.Locals {
-		if err = f.Locals[i].UnmarshalWASM(bytesReader); err != nil {
+	for i := 0; i < int(localCount); i++ {
```
sbinet:
```suggestion
	for i := uint32(0); i < localCount; i++ {
```

wasm/section.go
```diff
@@ -849,12 +863,14 @@ func (f *FunctionBody) UnmarshalWASM(r io.Reader) error {
...
+		local := LocalEntry{}
```
sbinet:
```suggestion
		var local LocalEntry
```

wasm/section.go
```diff
@@ -934,11 +950,13 @@ func (s *SectionData) ReadPayload(r io.Reader) error {
 	if err != nil {
 		return err
 	}
-	s.Entries = make([]DataSegment, count)
-	for i := range s.Entries {
-		if err = s.Entries[i].UnmarshalWASM(r); err != nil {
+	s.Entries = make([]DataSegment, 0, getInitialCap(count))
+	for i := 0; i < int(count); i++ {
```
sbinet:
```suggestion
	for i := uint32(0); i < count; i++ {
```

wasm/section.go
```diff
@@ -934,11 +950,13 @@ func (s *SectionData) ReadPayload(r io.Reader) error {
...
+		entry := DataSegment{}
```
sbinet:
```suggestion
		var entry DataSegment
```

wasm/types.go
```diff
@@ -134,26 +134,28 @@ func (f *FunctionSig) UnmarshalWASM(r io.Reader) error {
 	if err != nil {
 		return err
 	}
-	f.ParamTypes = make([]ValueType, paramCount)
+	f.ParamTypes = make([]ValueType, 0, getInitialCap(paramCount))
 
-	for i := range f.ParamTypes {
-		err = f.ParamTypes[i].UnmarshalWASM(r)
-		if err != nil {
+	for i := 0; i < int(paramCount); i++ {
```
sbinet:
```suggestion
	for i := uint32(0); i < paramCount; i++ {
```

wasm/types.go
```diff
@@ -134,26 +134,28 @@ func (f *FunctionSig) UnmarshalWASM(r io.Reader) error {
...
+		val := ValueType(0)
```
sbinet:
```suggestion
		var v ValueType
```

wasm/types.go
```diff
@@ -134,26 +134,28 @@ func (f *FunctionSig) UnmarshalWASM(r io.Reader) error {
...
+		if err = val.UnmarshalWASM(r); err != nil {
```
sbinet:
```suggestion
		if err = v.UnmarshalWASM(r); err != nil {
```

wasm/types.go
```diff
@@ -134,26 +134,28 @@ func (f *FunctionSig) UnmarshalWASM(r io.Reader) error {
...
 	if err != nil {
 		return err
 	}
 
-	f.ReturnTypes = make([]ValueType, returnCount)
-	for i := range f.ReturnTypes {
-		err = f.ReturnTypes[i].UnmarshalWASM(r)
-		if err != nil {
+	f.ReturnTypes = make([]ValueType, 0, getInitialCap(returnCount))
+	for i := 0; i < int(returnCount); i++ {
```
sbinet:
```suggestion
	for i := uint32(0); i < returnCount; i++ {
```

wasm/types.go
```diff
@@ -134,26 +134,28 @@ func (f *FunctionSig) UnmarshalWASM(r io.Reader) error {
...
+		val := ValueType(0)
```
sbinet:
```suggestion
		var v ValueType
```

wasm/types.go
```diff
@@ -134,26 +134,28 @@ func (f *FunctionSig) UnmarshalWASM(r io.Reader) error {
...
+		if err = val.UnmarshalWASM(r); err != nil {
```
sbinet:
```suggestion
		if err = v.UnmarshalWASM(r); err != nil {
```

wasm/types.go
```diff
@@ -134,26 +134,28 @@ func (f *FunctionSig) UnmarshalWASM(r io.Reader) error {
...
 			return err
 		}
+		f.ReturnTypes = append(f.ReturnTypes, val)
```
sbinet:
```suggestion
		f.ReturnTypes = append(f.ReturnTypes, v)
```

wasm/types.go
```diff
@@ -134,26 +134,28 @@ func (f *FunctionSig) UnmarshalWASM(r io.Reader) error {
...
 
-	for i := range f.ParamTypes {
-		err = f.ParamTypes[i].UnmarshalWASM(r)
-		if err != nil {
+	for i := 0; i < int(paramCount); i++ {
+		val := ValueType(0)
+		if err = val.UnmarshalWASM(r); err != nil {
 			return err
 		}
+		f.ParamTypes = append(f.ParamTypes, val)
```
sbinet:
```suggestion
		f.ParamTypes = append(f.ParamTypes, v)
```

PR #149, opened 2019-07-30
exec: expose memory size to Process

laizy:
we current has some host function like:
```golang
func HostFunc(proc *exec.Process, keyPtr uint32, keylen uint32) {
	keybytes := make([]byte, keylen) 
	_, err := proc.ReadAt(keybytes, int64(keyPtr))
       ...
}
```
vicious wasm code can provide a large `keylen` to make the bytes allocation failed. Expose the current memory size can let us prevent this attack by checking `keyPtr + keylen <= MemSize`


exec/vm.go
```diff
@@ -531,6 +531,11 @@ func (proc *Process) WriteAt(p []byte, off int64) (int, error) {
 	return length, err
 }
 
+// MemSize returns the current allocated memory size in bytes
```
sbinet: `s/in bytes/in bytes./`

laizy: fixed

PR #150, opened 2019-07-31
wasm: fix index out of range panic

laizy:
The large_table.wat file passed the `wasm-validate` check, but will cause panic in wagon.

Wasm spec says: 

> Integers are not inherently signed or unsigned, their interpretation is determined by individual operations.

So the `offset` should be treated as `uint32`.


wasm/index.go
```diff
@@ -126,28 +126,29 @@ func (m *Module) populateTables() error {
 	for _, elem := range m.Elements.Entries {
 		// the MVP dictates that index should always be zero, we should
 		// probably check this
-		if int(elem.Index) >= len(m.TableIndexSpace) {
+		if elem.Index >= uint32(len(m.TableIndexSpace)) {
```
laizy: In 32bit system, `int` may only have 32 bits, so if `elem.Index` is too large, `int(elem.Index)` may be negative, and pass through this test, cause panic in the following operation.

laizy:
test case passed in my local machine, seems travis-ci has memory limitation, so I will remove the file from testdata, and paste the content here as a reference:
```wasm
(module
(table 1 anyfunc)
(elem 0 (i32.const 2147483648) 0)
(func (result i32)
(i32.const 10)
)
(export "main" (func 0))
)
```

PR #151, opened 2019-08-01
wasm: check sections order

### LINKED ISSUE #144: function DecodeModule implementation is not consistent with the wasm specification
from the [WebAssembly Specifications](https://webassembly.github.io/spec/core/binary/modules.html#binary-importsec)
> The preamble is followed by a sequence of sections. Custom sections may be inserted at any place in this sequence, while other sections must occur at most once and in the prescribed order. All sections can be empty.

the current implementation use a [for loop](https://github.com/go-interpreter/wagon/blob/master/wasm/module.go#L120) to decode the section data. so the prescribed order is not checked. 

A similar processing logic as a reference is at https://github.com/CraneStation/cranelift/blob/master/cranelift-wasm/src/module_translator.rs#L14

laizy:
fix #144


wasm/section.go
```diff
@@ -105,19 +105,45 @@ func (e MissingSectionError) Error() string {
...
 		return false, err
 	}
+	if id != uint8(SectionIDCustom) {
+		if id <= sr.lastSecOrder {
+			return false, fmt.Errorf("wasm: sections must occur at most once and in the prescribed order")
+		}
+		sr.lastSecOrder = id
+	}
+
 	s := RawSection{ID: SectionID(id)}
```
laizy: reading out the section id as uint32 and truncating to uint8 will cause problem.

wasm/module.go
```diff
@@ -116,15 +117,17 @@ func DecodeModule(r io.Reader) (*Module, error) {
...
 
-	for {
-		done, err := m.readSection(reader)
-		if err != nil {
-			return nil, err
-		} else if done {
-			return m, nil
-		}
+	secReader := newSectionsReader(m)
+	err = secReader.readSections(reader)
```
sbinet: no need to really name that variable, is it?
what about just:
```go
err = newSectionsReader(m).readSections(reader)
```

wasm/section.go
```diff
@@ -105,19 +105,45 @@ func (e MissingSectionError) Error() string {
...
+}
+
+func newSectionsReader(m *Module) sectionsReader {
+	return sectionsReader{m: m}
+}
+
+func (s *sectionsReader) readSections(r *readpos.ReadPos) error {
+	for {
+		done, err := s.readSection(r)
+		if err != nil {
```
sbinet: I'd use a `switch` like so:
```go
switch {
case err != nil:
    return err
case done:
    return nil
}
```

wasm/section.go
```diff
@@ -105,19 +105,46 @@ func (e MissingSectionError) Error() string {
 	return fmt.Sprintf("wasm: missing section %s", SectionID(e).String())
 }
 
+type sectionsReader struct {
+	lastSecOrder uint8 // previous non-custom sectionid
+	m            *Module
+}
+
+func newSectionsReader(m *Module) sectionsReader {
```
sbinet:
```suggestion
func newSectionsReader(m *Module) *sectionsReader {
```

wasm/section.go
```diff
@@ -105,19 +105,46 @@ func (e MissingSectionError) Error() string {
...
+	return sectionsReader{m: m}
```
sbinet:
```suggestion
	return &sectionsReader{m: m}
```

PR #152, opened 2019-08-02
wasm: fix types decoding

laizy:


wasm/types.go
```diff
@@ -22,13 +23,13 @@ type Unmarshaler interface {
...
 
 const (
-	ValueTypeI32 ValueType = -0x01
-	ValueTypeI64 ValueType = -0x02
-	ValueTypeF32 ValueType = -0x03
-	ValueTypeF64 ValueType = -0x04
+	ValueTypeI32 ValueType = 0x7f
+	ValueTypeI64 ValueType = 0x7e
+	ValueTypeF32 ValueType = 0x7d
+	ValueTypeF64 ValueType = 0x7c
```
laizy: These constants are obtained from the specification: https://webassembly.github.io/spec/core/binary/types.html#binary-valtype

wasm/read.go
```diff
@@ -35,6 +35,17 @@ func readBytes(r io.Reader, n int) ([]byte, error) {
 	return nil, io.ErrUnexpectedEOF
 }
 
+func WriteByte(w io.Writer, b byte) error {
```
sbinet: perhaps put it in a new `wasm/write.go` file?

also, perhaps:
```suggestion
func writeByte(w io.Writer, b byte) error {
```

wasm/types.go
```diff
@@ -59,13 +60,13 @@ func (t *ValueType) UnmarshalWASM(r io.Reader) error {
 }
 
 func (t ValueType) MarshalWASM(w io.Writer) error {
-	_, err := leb128.WriteVarint64(w, int64(t))
+	err := WriteByte(w, byte(t))
```
sbinet:
```suggestion
	return WriteByte(w, byte(t))
```

wasm/read.go
```diff
@@ -35,6 +35,17 @@ func readBytes(r io.Reader, n int) ([]byte, error) {
...
+	_, err := w.Write([]byte{b})
+	return err
+}
+
+func ReadByte(r io.Reader) (byte, error) {
```
sbinet: perhaps not export it?
```suggestion
func readByte(r io.Reader) (byte, error) {
```

laizy: The `ReadByte` method is used in `disasm` package.

wasm/read.go
laizy: The `ReadByte` method is used in `disasm` package. For the sake of symmetry, I exposed it too. and maybe it will be used in later fix.

laizy: fixed

wasm/read.go
sbinet: ah. right.

PR #153, opened 2019-08-05
wasm: checks names are valid utf-8 string

### LINKED ISSUE #145: names used in wasm import/export/module etc need to be valid utf-8 string
from the [WebAssembly Specifications](https://webassembly.github.io/spec/core/bikeshed/index.html#names%E2%91%A0):
> Names are sequences of characters, which are scalar values as defined by [UNICODE] (Section 2.4).

the golang string is just raw bytes wrapper, so need add utf-8 encoding check at https://github.com/go-interpreter/wagon/blob/master/wasm/read.go#L32

laizy:
fix #145

### DIFF
--- a/wasm/read.go
+++ b/wasm/read.go
@@ -56,7 +56,7 @@ func readBytesUint(r io.Reader) ([]byte, error) {
 	return readBytes(r, n)
 }
 
-func readUtf8String(r io.Reader, n uint32) (string, error) {
+func readUTF8String(r io.Reader, n uint32) (string, error) {
 	bytes, err := readBytes(r, n)
 	if err != nil {
 		return "", err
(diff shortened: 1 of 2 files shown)


wasm/read.go
```diff
@@ -51,23 +53,26 @@ func readBytesUint(r io.Reader) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-	return readBytes(r, int(n))
+	return readBytes(r, n)
 }
 
-func readString(r io.Reader, n int) (string, error) {
+func readUtf8String(r io.Reader, n uint32) (string, error) {
```
sbinet: `s/Utf8/UTF8/`

wasm/read.go
```diff
@@ -51,23 +53,26 @@ func readBytesUint(r io.Reader) ([]byte, error) {
...
 		return "", err
 	}
+	if !utf8.Valid(bytes) {
+		return "", errors.New("wasm: invalid utf-8 string")
+	}
 	return string(bytes), nil
 }
 
-func readStringUint(r io.Reader) (string, error) {
+func readUtf8StringUint(r io.Reader) (string, error) {
```
sbinet: `s/Utf8/UTF8/`

PR #155, opened 2019-08-08
wasm: fix leb128 decoding

laizy:
the original implementation missing some checks:
1. overflow check
2. encoded integer must not exceed ceil(N/7) bytes.
3. `ReadVarint32` is directly truncated from `int64`


laizy:
travis-ci seems unstable.

wasm/leb128/read.go
```diff
@@ -7,66 +7,81 @@
...
-	var (
-		b     = make([]byte, 1)
-		shift uint
-		res   uint32
-		err   error
-	)
-	for {
-		if _, err = io.ReadFull(r, b); err != nil {
-			return res, err
+// read uN defined in https://webassembly.github.io/spec/core/binary/values.html#binary-int
```
sbinet:
```suggestion
// readVarUint reads an unsigned integer of size n defined in https://webassembly.github.io/spec/core/binary/values.html#binary-int
```

wasm/leb128/read.go
```diff
@@ -7,66 +7,81 @@
...
+// note: the result type is uint64, so N must <= 64.
```
sbinet: we should make sure that `n<=64` and:
```suggestion
// readVarUint panics if n>64.
```

wasm/leb128/read.go
```diff
@@ -7,66 +7,81 @@
...
+func readVarUint(r io.Reader, N uint) (uint64, error) {
```
sbinet:
```suggestion
func readVarUint(r io.Reader, n uint) (uint64, error) {
```

wasm/leb128/read.go
```diff
@@ -7,66 +7,81 @@
...
+	p := make([]byte, 1)
+	_, err := r.Read(p)
```
sbinet: we should probably use `io.ReadFull` instead of just an `io.Reader.Read` to properly handle `wasm` files read over the network (or any non-completely reliable media).

wasm/leb128/read.go
```diff
@@ -7,66 +7,81 @@
...
+	if err != nil {
+		return 0, err
+	}
+	n := uint64(p[0])
```
sbinet:
```suggestion
	nn := uint64(p[0])
```

wasm/leb128/read.go
```diff
@@ -7,66 +7,81 @@
...
+	if n < 1<<7 && n < 1<<N {
```
sbinet: I'd rather use a `switch` instead of an `if/else-if`.

wasm/leb128/read.go
```diff
@@ -7,66 +7,81 @@
...
-		res |= (cur & 0x7f) << (shift)
-		if cur&0x80 == 0 {
-			return res, nil
+		return (1<<7)*m + (n - 1<<7), nil
+	}
+
+	return 0, errors.New("leb128: invalid uint")
+}
+
+// read sN defined in https://webassembly.github.io/spec/core/binary/values.html#binary-int
```
sbinet:
```suggestion
// readVarint reads a signed integer of size n, defined in https://webassembly.github.io/spec/core/binary/values.html#binary-int
```

(same modification for the `note` below.)

wasm/leb128/read.go
```diff
@@ -7,66 +7,81 @@
...
-		res   uint32
-		err   error
-	)
-	for {
-		if _, err = io.ReadFull(r, b); err != nil {
-			return res, err
+// read uN defined in https://webassembly.github.io/spec/core/binary/values.html#binary-int
+// note: the result type is uint64, so N must <= 64.
+func readVarUint(r io.Reader, N uint) (uint64, error) {
+	p := make([]byte, 1)
```
sbinet: add a check+panic on `n` before this line?

wasm/leb128/read_test.go
```diff
@@ -70,3 +84,67 @@ func TestReadVarint32Err(t *testing.T) {
...
+		if err != nil {
+			continue
+		}
+		readLen := len(buf) - reader.Len()
+		assert.True(t, readLen <= (64+6)/7) // ceil(N/7) bytes
+
+		buf2 := bytes.NewBuffer(nil)
+		WriteVarint64(buf2, val)
+		if readLen <= len(buf2.Bytes()) {
+			assert.True(t, bytes.HasPrefix(buf, buf2.Bytes()), fmt.Sprintf("val:%d, origin buf:%v, buf2: %v", val, buf, buf2.Bytes()))
```
sbinet: please don't use `testify/assert` and just use standard Go constructs.

wasm/leb128/read_test.go
```diff
@@ -70,3 +84,67 @@ func TestReadVarint32Err(t *testing.T) {
...
+		if err != nil {
+			continue
+		}
+		readLen := len(buf) - reader.Len()
+		assert.True(t, readLen <= (32+6)/7) // ceil(N/7) bytes
+
+		buf2 := bytes.NewBuffer(nil)
+		WriteVarUint32(buf2, val)
+		if readLen <= len(buf2.Bytes()) {
+			assert.True(t, bytes.HasPrefix(buf, buf2.Bytes()), fmt.Sprintf("val:%d, origin buf:%v, buf2: %v", val, buf, buf2.Bytes()))
```
sbinet: ditto.

wasm/leb128/read.go
```diff
@@ -7,66 +7,81 @@
...
+	_, err := r.Read(p)
+	if err != nil {
+		return 0, err
+	}
+	n := uint64(p[0])
+	if n < 1<<7 && n < 1<<N {
+		return n, nil
+	} else if n >= 1<<7 && N > 7 {
+		m, err := readVarUint(r, N-7)
```
sbinet: the original implementation was using a loop.
now we are using recursion (at which Go isn't particularly good at), plus we allocate a 1-len slice everytime.

couldn't we switch back to a loop?
or at least pass a "work" slice as argument to `readVarUint`?

laizy: I have implemented a loop version.
the specification format has some differences with https://en.wikipedia.org/wiki/LEB128.
allowing trialing zeros, but having upper bound encoding size makes it hard to get right.
this recursion version is pretty close to the formula definition of the specification. so I left it as a baseline to test the loop one.

wasm/leb128/read.go
```diff
@@ -7,66 +7,146 @@
...
 	}
 }
 
-// ReadVarint32 reads a LEB128 encoded signed 32-bit integer from r, and
-// returns the integer value, and the error (if any).
-func ReadVarint32(r io.Reader) (int32, error) {
-	n, err := ReadVarint64(r)
-	return int32(n), err
-}
+func readVarUintRecur(r io.Reader, n uint) (uint64, error) {
```
sbinet: please move this function to the `_test.go` file (so we don't use it unadvertantly).

wasm/leb128/read.go
```diff
@@ -7,66 +7,146 @@
...
+	}
+}
 
-		cur := int64(b[0])
-		res |= (cur & 0x7f) << shift
-		shift += 7
-		sign <<= 7
-		if cur&0x80 == 0 {
-			break
+func readVarintRecur(r io.Reader, n uint) (int64, error) {
```
sbinet: ditto.

PR #156, opened 2019-08-08
validate/disasm: fix call_indirect and some reserved value checking

laizy:


disasm/disasm.go
```diff
@@ -428,10 +428,13 @@ func Disassemble(code []byte) ([]Instr, error) {
 			}
 			instr.Immediates = append(instr.Immediates, index)
 			if op == ops.CallIndirect {
-				reserved, err := leb128.ReadVarUint32(reader)
+				reserved, err := wasm.ReadByte(reader)
```
sbinet:
```suggestion
				idx, err := wasm.ReadByte(reader)
```

disasm/disasm.go
```diff
@@ -428,10 +428,13 @@ func Disassemble(code []byte) ([]Instr, error) {
...
 				if err != nil {
 					return nil, err
 				}
+				if reserved != 0x00 {
```
sbinet:
```suggestion
				if idx != 0x00 {
```

disasm/disasm.go
```diff
@@ -468,22 +471,25 @@ func Disassemble(code []byte) ([]Instr, error) {
...
+			instr.Immediates = append(instr.Immediates, align)
 
 			offset, err := leb128.ReadVarUint32(reader)
 			if err != nil {
 				return nil, err
 			}
 			instr.Immediates = append(instr.Immediates, offset)
 		case ops.CurrentMemory, ops.GrowMemory:
-			res, err := leb128.ReadVarUint32(reader)
+			res, err := wasm.ReadByte(reader)
```
sbinet:
```suggestion
			idx, err := wasm.ReadByte(reader)
```

validate/validate.go
```diff
@@ -298,6 +302,17 @@ func verifyBody(fn *wasm.FunctionSig, body *wasm.FunctionBody, module *wasm.Modu
...
 			}
+			tableIndex, err := vm.fetchByte()
+			if err != nil {
+				return vm, err
+			}
+			if tableIndex != 0x00 {
+				return vm, errors.New("validate: table index in call_indirect must be 0")
+			}
+
+			if index >= uint32(len(module.Types.Entries)) {
```
sbinet: perhaps move this check closer to the type index value extraction? (ie: to line 305)

laizy: I casually added this, since I found `index` used in line 317 is not bound checked.  so it is reasonable to place this check here just before the place of use?

sbinet: that stands to reason as well.

PR #166, opened 2019-09-13
Fixes panic on terminate for return values

meyer9:
If you call `proc.Terminate()` in an external function, sometimes the VM will expect a return value on the stack even though there isn't one there because the function didn't finish. In this case, we should ignore the return value from the function.

### DIFF
--- a/exec/vm.go
+++ b/exec/vm.go
@@ -452,7 +452,7 @@ outer:
 		}
 	}
 
-	if compiled.returns {
+	if compiled.returns && !vm.abort {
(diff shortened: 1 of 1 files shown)


sbinet:
Thanks for the PR.
Could you send another one against the go-interpreter/license repo adding yourself to the authors& contributors files?
Then I'll merge everything.

Thanks.

sbinet:
Ping?

meyer9:
Doing that now, sorry missed your comment.

sbinet:
thanks :)
(and no worries)

PR #172, opened 2019-10-09
Expose access to Ops to avoid stringly typing downstream

silasdavis:
Signed-off-by: Silas Davis <silas@monax.io>


wasm/operators/op.go
```diff
@@ -85,3 +85,7 @@ func New(code byte) (Op, error) {
 	}
 	return op, nil
 }
+
+func Get(opcode byte) Op {
```
twitchyliquid64: Could you add a doc string for this?

iwasaki-kenta:
Might using `Stringer` for op's be a better option over directly fetching the name of the op?

https://github.com/golang/tools/blob/master/cmd/stringer/stringer.go

twitchyliquid64:
Feel free to integrate stringer for `Op` if you wish.

Apologies for the confusion, what I meant was to add a doc comment to `Get(opcode byte) Op`. Its part of the exported API surface, so we need a doc comment so it shows up with documentation in godoc (See the Commentary --> Doc comments section in [effective Go](https://golang.org/doc/effective_go.html#commentary) for more rationale).

sbinet:
_ping?_

PR #176, opened 2019-10-29
leb128:add ReadVarUint64 function and fix overflow corner case

### LINKED ISSUE #175: Read leb128 fail when trying to read 64 bit of small number
```golang
r := bytes.NewReader([]byte{0x08})
n, err := ReadUint64(r)
if err != nil {
  t.Fatal(err)
}
fmt.Println(n)
// output: leb128: invalid uint
```

laizy:
Close #175


wasm/leb128/read_test.go
```diff
@@ -217,6 +234,8 @@ func readVarUintRecur(r io.Reader, n uint) (uint64, error) {
 	}
 }
 
+// a less efficent recurrsive version but matches the spec defined in https://webassembly.github.io/spec/core/binary/values.html#integers
```
twitchyliquid64: s/recurrsive/recursive

laizy:
```suggestion
// a less efficent recursive version but matches the spec defined in https://webassembly.github.io/spec/core/binary/values.html#integers
```

wasm/leb128/read_test.go
```diff
@@ -192,6 +206,8 @@ func TestCompareReadVarUint(t *testing.T) {
...
```
laizy:
```suggestion
// a less efficent recursive version but matches the spec defined in https://webassembly.github.io/spec/core/binary/values.html#integers
```

PR #178, opened 2019-11-13
Fix type conversion panic for host-provided funcs

ericflo:
It was panicking because it's of type goFunction, not *goFunction as the line is checking for.

### DIFF
--- a/exec/native_compile.go
+++ b/exec/native_compile.go
@@ -91,7 +91,7 @@ func (vm *VM) tryNativeCompile() error {
 	}
 
 	for i := range vm.funcs {
-		if _, isGoFunc := vm.funcs[i].(*goFunction); isGoFunc {
+		if _, isGoFunc := vm.funcs[i].(goFunction); isGoFunc {
 			continue
 		}
 


exec/native_compile.go
```diff
@@ -91,7 +91,7 @@ func (vm *VM) tryNativeCompile() error {
 	}
 
 	for i := range vm.funcs {
-		if _, isGoFunc := vm.funcs[i].(*goFunction); isGoFunc {
+		if _, isGoFunc := vm.funcs[i].(goFunction); isGoFunc {
```
sbinet: hum...
this kind of type assertion, using the "comma-ok" variant, shouldn't panic.
so I must admit I am not completely clear on what `panic` this should cure.

that said, it seems indeed that we should probably type-assert for `goFunction`:
```
$> git grep goFunction
exec/func.go:38:type goFunction struct {
exec/func.go:43:func (fn goFunction) call(vm *VM, index int64) {
exec/native_compile.go:94:              if _, isGoFunc := vm.funcs[i].(*goFunction); isGoFunc {
exec/native_compile.go:181:             if _, isGoFunc := vm.funcs[i].(*goFunction); isGoFunc {
exec/vm.go:130:                 vm.funcs[i] = goFunction{
```
(as we only ever fill `vm.funcs` with `goFunction{...}` values...)

sbinet (commented):
thanks for the PR.

could you add a test/example showing the panic you are trying to fix with this PR?

thanks again.

exec/native_compile.go
ericflo: Sorry, my fault for underspecifying - it passes this check, so in the code below it (line 98) it panics trying to convert access it as a compiledFunction

ericflo:
I'll look into adding a new test, but one way to test it would be to run the existing host-provided func test with AOT enabled

exec/native_compile.go
twitchyliquid64: Hmmm, I dont understand how we would hit this panic. Consider that `vm.funcs` is of type `[]function`, and `function` is an interface type. The `goFunction` type only implements the necessary interface methods on itself, it does not implement it on its pointer receiver.

As such, I was thinking only `goFunction{}` could implement `function`, not a `*goFunction` ? Now I'm not sure ...

Either way, the current logic is negated, considering we only want to run the loop logic on the func if it is of type `compiledFunction`.

What do you think of a fix like this?

```go
fn, isCompiledFn := vm.funcs[i].(compiledFunction)
if !isCompiledFn {
  continue
}
```

sbinet:
ping?

fabian-z:
I can confirm that using host provided functions with the AOT backend enabled causes a panic for me that is fixed by this change (I tested with rebasing on top of master).

The panic happens here:
https://github.com/go-interpreter/wagon/blob/dc6758ca7262efb2ef987e32e09f648300b4d007/exec/native_compile.go#L98

Stack trace:
```
panic: interface conversion: exec.function is exec.goFunction, not exec.compiledFunction

goroutine 1 [running]:
github.com/go-interpreter/wagon/exec.(*VM).tryNativeCompile(0xc0007fc000, 0xc00006f770, 0x0)
	/home/fabian/go/pkg/mod/github.com/go-interpreter/wagon@v0.6.1-0.20200313144839-dc6758ca7262/exec/native_compile.go:98 +0x85d
github.com/go-interpreter/wagon/exec.NewVM(0xc00013c000, 0xc0007e7f30, 0x1, 0x1, 0x0, 0x0, 0x0)
	/home/fabian/go/pkg/mod/github.com/go-interpreter/wagon@v0.6.1-0.20200313144839-dc6758ca7262/exec/vm.go:175 +0x4cd
main.main()
	/home/fabian/playground/wasm-runtime/main.go:285 +0x1a8

```

PR #184, opened 2020-02-01
Fix shift instructions

zxh0:
Hi, as the [Wasm Spec](https://webassembly.github.io/spec/core/exec/numerics.html#op-ishl) said, all shift instructions should mod the second arg: `Let k be i2 modulo N.` . And [here](https://webassembly.studio/?f=qn4yobte40g) is a simple online test case. I have fixed this bug and added some unit tests. Please double check this.

### DIFF
--- a/exec/num.go
+++ b/exec/num.go
@@ -76,19 +76,19 @@ func (vm *VM) i32Xor() {
 func (vm *VM) i32Shl() {
 	v2 := vm.popUint32()
 	v1 := vm.popUint32()
-	vm.pushUint32(v1 << v2)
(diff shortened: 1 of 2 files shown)


laizy (approved):

laizy:
this patch passed the shift instructions testcases in file `i32.wast`, `i64.wast`  in  https://github.com/WebAssembly/testsuite.

twitchyliquid64:
Did the AOT compiler need any changes or it just happened to be correct (luckily) ?

sbinet:
ping?

PR #189, opened 2020-02-26
wasm: use io.ReadFull instead of Read

### LINKED ISSUE #188: io.ReadFull should be used at ReadByte
https://github.com/go-interpreter/wagon/blob/v0.6.0/wasm/read.go#L45

In theory, `Read` might return 0 bytes without errors. `io.ReadFull` should be used to ensure that 1 byte is read.

hajimehoshi:
There is no guarantee that io.Read fulfills the given byte slice.
ReadByte should use io.ReadFull to ensure that instead.

Fixes #188


hajimehoshi:
https://github.com/go-interpreter/license/pull/20 done :-)

Issue #190, opened 2020-02-27
wasm: functions in FunctionIndexSpace include wrong contents


hajimehoshi:
1. Build a hello-world wasm:

```go
package main

func main() {
        println("Hello, World!")
}
```

```
GOOS=js GOARCH=wasm go build -o test.wasm helloworld.go 
```

2. Run the below code:

```go
package main

import (
        "fmt"
        "os"

        "github.com/go-interpreter/wagon/wasm"
)

func main() {
        f, err := os.Open("test.wasm")
        if err != nil {
                panic(err)
        }
        defer f.Close()

        mod, err := wasm.ReadModule(f, nil)
        if err != nil {
                panic(err)
        }

        fn := mod.FunctionIndexSpace[0]
        fmt.Printf("name: %q\n", fn.Name) // This is empty. That's fine since this is an imported function 'debug'.                                                                                                                           
        fmt.Printf("len(code): %d\n", len(fn.Body.Code)) // This is not empty. That's unexpected. This content is for the first function as 'go.buildid'.                                                                                     
}
```

It looks like the function's index and its instructions don't match. Some of the first functions should be imported functions, but include instructions for other functions. I think the function bodies are shifted.

hajimehoshi:
https://github.com/go-interpreter/wagon/blob/cbbdf0893ce4f79668a13c0879a7f143bdd0f472/wasm/index.go#L76-L80

The names are shifted with `numImports`, which is correct, while Sig and Body are not shifted.

zhangzhiqiangcs:
Hello, I read in detail, found that the problem comes with numImports, when you don't provide a imported module but the wasm file need it, the numImports is not equal to len(FunctionIndexSpace).
https://github.com/go-interpreter/wagon/blob/dc6758ca7262efb2ef987e32e09f648300b4d007/wasm/index.go#L84-L85
So I had a pr #193 
I am not sure if we can do it in this way.
Thanks.

PR #193, opened 2020-04-14
calculate number of imported funcs

zhangzhiqiangcs:
Hi, I read the bug in #190 
The problem is that when providing a nil imported object, the len(FunctionIndexSpace) is not equal to the number of imported funcs. 
So I calculate it. I don't know if we can do it this way.
Thanks.


wasm/index.go
```diff
@@ -82,7 +82,17 @@ func (m *Module) populateFunctions() error {
 	}
 
 	// Add the functions from the wasm itself to the function list
-	numImports := len(m.FunctionIndexSpace)
+
+	// when imported module not provided, len(m.FunctionIndecSpace) == 0
```
sbinet, line 86: could you reword a bit this comment?
I am not sure I completely understand it.

do you mean:
```
// when an imported module is not provided, the functions index space is empty.
// to maintain indices consistency, we compute the number of expected imported functions.
```

sbinet (commented):
thanks.

could you add a small test for this?

wasm/index.go
zhangzhiqiangcs: yes, that's it.

